PR #32, opened 2020-01-01
Handle visibility and docblocks of constants

simivar:
|    Q        |   A
|------------ | ------
| Bugfix      | no
| BC Break    | yes
| New Feature | yes
| RFC         | yes

### Description
Currently library does not handle visibility of `const` fields in classes at all. All generated `const`s are defined as `public`. In this PR I've added third argument to `addConstant` method the same way that it is defined in `addProperty` method which User can use to set visibility of `const`. 

Also, I've created new `ConstantReflection` that handles generating `const` from reflection with all data that it had in original class: visibility, dockblocks. 

~In order to keep it as BC~ I introduced `FieldsReflectionTrait` and `FieldsReflectionInterface` that is used and implemented by both `ConstantReflection` and `PropertyReflection`. I tried to come up with different, better solution as this one seems not good enough for me but couldn't. I am open to all suggestions and I will gladly make necessary changes.

### Example 1
```php
<?php

require_once 'vendor/autoload.php';

use Laminas\Code\Generator\ClassGenerator;
use Laminas\Code\Reflection\ClassReflection;

class Test {
    /**
     * Description.
     *
     * @var string Second description.
     */
    protected const A = 'b';
};

$generator = ClassGenerator::fromReflection(
    new ClassReflection(Test::class)
);

$code = $generator->generate();
echo $code;
```

Before changes:
```php
class Test
{

    public const A = 'b';


}
```

After changes:
```php
class Test                                
{                                         
                                          
    /**                                   
     * Description.                       
     *                                    
     * @var string Second description.    
     */                                   
    protected const A = 'b';              
                                          
                                          
}
```

### Example 2
```php
use Laminas\Code\Generator\ClassGenerator;
use Laminas\Code\Generator\PropertyGenerator;

$foo = new ClassGenerator();
$foo->setName('Foo')
    ->addConstant('A', 'b', PropertyGenerator::FLAG_PROTECTED);

echo $foo->generate();
```

results in
```php
class Foo
{

    protected const A = 'b';


}
```


michalbundyra:
@simivar Just looked quickly and this is definitely BC Break. I don't think we can do it without and keep BC. Adding parameter to public method in non final class or changing parameter type hint are BC breaks.

simivar:
@michalbundyra oh, yeah, you’re right. I got so caught up to keep it BC that I didn’t notice when I broke BC. I will think of some other way then using ‘trait’ to make that change. I am still open to all suggestions how you feel it should be done.

Ocramius:
Closing here: this patch is pre-`Laminas` rename, from what I can see, which means that the diff ended up being ginormous (due to foreign commits being in it).

The usage of a trait here is probably to be avoided, but I'd need to review a cleaned up patch, after evaluating that.

If interested in pursuing this further, I suggest re-opening a patch on top of `4.6.x`, with just these commits in it:

 * https://github.com/laminas/laminas-code/pull/32/commits/ae7d71fa374365e2ce16297a63cd4aeea32d8cb4
 * https://github.com/laminas/laminas-code/pull/32/commits/2fbe9604b87bfeb9effe34d14a1184b7b256240f

PR #39, opened 2020-04-20
Suggest "laminas/laminas-zendframework-bridge"

olvlvl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Hi everyone,

With this PR, `laminas/laminas-zendframework-bridge` becomes a suggestion rather than a requirement.

We're trying to keep our bootstrap time as small as possible and this is not helping. One of our dependency uses `ocramius/proxy-manager`, which uses `laminas-code`, but we're not using Zend Framework.

Right now, we added this to our `composer.json`, but of course, a cleaner solution would be welcome:

```json
    "replace": {
        "laminas/laminas-zendframework-bridge": "^1.0"
    },
```


weierophinney:
Unfortunately, we cannot do this until the next major release.

In order for our component packages to act as **replacements** of the original ZF packages, users must be able to install them _in place of_ the original package, without otherwise changing their code. This means that if user code references a `Zend\Code` class, **that code must continue to work** after swapping packages.

The way we made this work was with the laminas-zendframework-bridge package, which adds autoloading rules to create `class_alias` entries for ZF classes that are now served by Laminas. Making the bridge a suggestion would cause the package to no longer work as a replacement, and is thus a BC break.

I'm going to leave this PR open, but have it target next major. I don't have a timeline for when we will do such a release, however. You will also need to rebase against develop, and change the target branch accordingly.

olvlvl:
I updated "BC Break" to "yes" and rebased against "develop". We'll keep our little hack for now.

Ocramius:
Manually merged via 765d88d37c95c0cd97710588f8d00dc927548dea (cherry-picked into a new branch)

PR #43, opened 2020-08-28
QA fix for phpstan

janvernieuwe:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

Small return type hint update to stop phpstan complaining.

It can return false when the array is empty, we check on that in the code, but because the type hint does not contain false, phpstan complains that the condition is always true (which it isn't in reality)

### DIFF
--- a/src/Generator/FileGenerator.php
+++ b/src/Generator/FileGenerator.php
@@ -10,6 +10,7 @@ namespace Laminas\Code\Generator;
 
 use Laminas\Code\DeclareStatement;
 use Laminas\Code\Exception\InvalidArgumentException;
+use Laminas\Code\Generator\Exception\ClassNotFoundException;
 use Laminas\Code\Reflection\Exception as ReflectionException;
 use Laminas\Code\Reflection\FileReflection;
 
@@ -335,18 +336,28 @@ class FileGenerator extends AbstractGenerator
     }
 
     /**
-     * @param  string $name
+     * @param string $name
+     *
      * @return ClassGenerator
+     * @throws ClassNotFoundException
      */
     public function getClass($name = null)
     {
         if ($name === null) {
             reset($this->classes);
+            $class = current($this->classes);
+            if (false === $class) {
+                throw new ClassNotFoundException('No class is set');
+            }
 
-            return current($this->classes);
+            return $class;
         }
 
-        return $this->classes[(string) $name];
+        if (false === array_key_exists($name, $this->classes)) {
+            throw new ClassNotFoundException(sprintf('Class %s is not set', $name));
+        }
+
+        return $this->classes[(string)$name];
     }
 
     /**
--- a/src/Generator/Exception/ClassNotFoundException.php
+++ b/src/Generator/Exception/ClassNotFoundException.php
@@ -1,0 +1,14 @@
+<?php
+
+/**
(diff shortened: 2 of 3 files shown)


src/Generator/FileGenerator.php
```diff
@@ -336,7 +336,7 @@ public function setClasses(array $classes)
 
     /**
      * @param  string $name
-     * @return ClassGenerator
+     * @return ClassGenerator|false
```
samsonasik: false is not data type, use `bool` instead

samsonasik:
Actually, we are going to use pslam, you can just rename the PR title

janvernieuwe:
> Actually, we are going to use pslam, you can just rename the PR title

We are using phpstan in our project, and it complains about it.
I'll change it to bool.

src/Generator/FileGenerator.php
Ocramius: `false` is acceptable since PHP 8 and in phpstan/psalm context, but we should really add an assertion instead of widening the type (which is a BC break)

Ocramius (changes requested):

src/Generator/FileGenerator.php
janvernieuwe: Well, i changed it to bool now.
But feel free to set it to false if you prefer that, either way it fixes my phpstan issue :)

src/Generator/FileGenerator.php
```diff
@@ -336,7 +336,7 @@ public function setClasses(array $classes)
 
     /**
      * @param  string $name
-     * @return ClassGenerator
+     * @return ClassGenerator|bool
```
Ocramius, line 339: As mentioned in comments below, a type change is a BC break - we should rather identify and fix this edge case scenario, by introducing a fallback or a crash.

Ocramius: As mentioned in comments before, a type change is a BC break - we should rather identify and fix this edge case scenario, by introducing a fallback or a crash.

Ocramius (changes requested):

src/Generator/FileGenerator.php
janvernieuwe:
```php
    /**
     * @param string $name
     *
     * @return ClassGenerator
     * @throws RuntimeException
     */
    public function getClass($name = null)
    {
        if ($name === null) {
            reset($this->classes);
            $class = current($this->classes);
            if (false === $class) {
                throw new RuntimeException('No class is set');
            }

            return $class;
        }

        if (false === array_key_exists($name, $this->classes)) {
            throw new RuntimeException(sprintf('Class %s is not set', $name));
        }

        return $this->classes[(string)$name];
    }
```

I'm thinking about something like this.
Would you prefer a more specific exception, like `ClassNotFoundException` (extending RuntimeException)?

Ocramius: If it is possible, a more precise exception is always preferable. It can extend `RuntimeException`, but it already makes everything much cleaner.

janvernieuwe: Alright, I added the checks and exception and switched base to develop.

janvernieuwe:
Any more changes needed or are we good here?

PR #47, opened 2020-10-21
Start PHP 8.0 support

### LINKED ISSUE #46: PHP 8.0 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

To be prepared for the december release of PHP 8.0, this repository has some additional TODOs to be tested against the new major version.

In order to make this repository compatible, one has to follow these steps:

- [ ] Modify `composer.json` to provide support for PHP 8.0 by adding the constraint `~8.0.0`
- [ ] Modify `composer.json` to drop support for PHP less than 7.3
- [ ] Modify `composer.json` to implement phpunit 9.3 which supports PHP 7.3+
- [ ] Modify `.travis.yml` to ignore platform requirements when installing composer dependencies (simply add `--ignore-platform-reqs` to `COMPOSER_ARGS` env variable)
- [ ] Modify `.travis.yml` to add PHP 8.0 to the matrix (*NOTE*: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [ ] Modify source code in case there are incompatibilities with PHP 8.0

Slamdunk:
Close https://github.com/laminas/laminas-code/issues/46

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,16 +21,16 @@
     "extra": {
     },
     "require": {
-        "php": "^7.1",
-        "laminas/laminas-eventmanager": "^2.6 || ^3.0",
-        "laminas/laminas-zendframework-bridge": "^1.0"
+        "php": "^7.3 || ~8.0.0",
+        "laminas/laminas-eventmanager": "^2.6 || ^3.3",
+        "laminas/laminas-zendframework-bridge": "^1.1"
     },
     "require-dev": {
         "ext-phar": "*",
-        "doctrine/annotations": "^1.7",
-        "laminas/laminas-coding-standard": "^1.0",
-        "laminas/laminas-stdlib": "^2.7 || ^3.0",
-        "phpunit/phpunit": "^7.5.16 || ^8.4"
+        "doctrine/annotations": "^1.10.4",
+        "laminas/laminas-coding-standard": "^1.0.0",
+        "laminas/laminas-stdlib": "^2.7.7 || ^3.3.0",
+        "phpunit/phpunit": "^9.4.2"
     },
     "conflict": {
         "phpspec/prophecy": "<1.9.0"
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require": {
         "php": "^7.3 || ~8.0.0",
-        "laminas/laminas-eventmanager": "^2.6 || ^3.3",
+        "laminas/laminas-eventmanager": "^3.3",
         "laminas/laminas-zendframework-bridge": "^1.1"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "ext-phar": "*",
         "doctrine/annotations": "^1.10.4",
         "laminas/laminas-coding-standard": "^1.0.0",
-        "laminas/laminas-stdlib": "^2.7.7 || ^3.3.0",
+        "laminas/laminas-stdlib": "^3.3.0",
         "phpunit/phpunit": "^9.4.2"
     },
     "conflict": {
--- a/src/Scanner/MethodScanner.php
+++ b/src/Scanner/MethodScanner.php
@@ -538,6 +538,8 @@ class MethodScanner implements ScannerInterface
 
             case T_VARIABLE:
             case T_STRING:
+            case T_NAME_QUALIFIED:
+            case T_NAME_FULLY_QUALIFIED:
                 if ($tokenType === T_STRING && $parentCount === 0) {
                     $this->name = $tokenContent;
                 }
--- a/src/Scanner/TokenArrayScanner.php
+++ b/src/Scanner/TokenArrayScanner.php
@@ -426,7 +426,12 @@ class TokenArrayScanner implements ScannerInterface
                     goto SCANNER_NAMESPACE_CONTINUE;
                 }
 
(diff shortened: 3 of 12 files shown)


Ocramius:
In order for this to work, we still need to make sure laminas-code does not crash with:

 * union types
 * `mixed` type
 * `static` return type
 * `throw` expression (`foo() ?? throw new \Exception();`)
 * `match` expression
 * null-safe expression (`?->`)
 * non-capturing catch (`try {} catch (\Exception) { return; }`)
 * named arguments calls
 * attributes
 * trailing comma in function arguments
 * variadic `ReflectionFunction::invoke()`, `ReflectionClass::newInstance()`, `ReflectionMethod::invoke()`
 * ctor property promotion

This is a lot of work that requires much more care for this specific package :S

Slamdunk:
If you agree, I'll do my best at least to start the support with the current test suite. I've renamed the PR accordingly: is it ok for you?

Ocramius:
@Slamdunk if you have time/willpower to throw at the problem, I can gladly support you! This has been on my to-do list as well, since `ocramius/proxy-manager` is highly dependant on `laminas/laminas-code`, and an upgrade to PHP 8 will be due soon there too :-)

Ocramius:
@Slamdunk an alternative (which had already be planned, but can't find where) is to drop all the scanner components from `laminas/laminas-code`: that's less effort, and I don't mind releasing a new major version of this package without that stuff.

That effectively means that everything that has to do with token changes can be removed, and therefore half the work is gone.

src/Scanner/MethodScanner.php
```diff
@@ -538,6 +538,8 @@ protected function scan()
 
             case T_VARIABLE:
             case T_STRING:
+            case T_NAME_QUALIFIED:
```
Slamdunk, line 541: Where should I put the backword compatibility definitions for the new constants?

Slamdunk:
@Ocramius well, I just made tests passing on PHP 8.0, I only need to make the BC, see https://github.com/laminas/laminas-code/pull/47#discussion_r509333954

src/Scanner/MethodScanner.php
Ocramius: Anywhere you want, as long as it's `@internal`, I suppose?

Slamdunk: Analyzing the source, `\Laminas\Code\Scanner\TokenArrayScanner::scan` seems the right place: https://github.com/laminas/laminas-code/pull/47/files#diff-c08ce1832ae9f8788d45e1cf8fa62d743ede600c44a4e622ec20b4b684383320R383
Had a run with `--process-isolation`, and everything works fine.

src/Scanner/TokenArrayScanner.php
```diff
@@ -380,6 +380,14 @@ protected function scan()
             return $infoIndex;
         };
 
+        // ensure php backwards compatibility
+        if (! defined('T_NAME_QUALIFIED')) {
+            define('T_NAME_QUALIFIED', 24000);
+        }
+        if (! defined('T_NAME_FULLY_QUALIFIED')) {
+            define('T_NAME_FULLY_QUALIFIED', 24000);
```
Slamdunk: Is there any particular value I should give to these constants?
For now I've just copy-pasted the one already present:
https://github.com/laminas/laminas-code/blob/8d74ede0d9b0850b4d7508997de48ed569151821/src/Scanner/ClassScanner.php#L1098-L1100

Slamdunk:
> if you have time/willpower to throw at the problem

From what you wrote, you clearly have much wider perspective on the overall future of this library.
Here I'm just trying to make everything run without fatal errors on PHP 8.0 as I'm upgrading my apps to be ready as soon as possible, so my work stops here for `laminas-code`.

May I ask you if this PR can be tagged after merge so I can continue my work on `laminas-form` (https://github.com/laminas/laminas-form/pull/75)?

composer.json
```diff
@@ -21,16 +21,16 @@
     "extra": {
     },
     "require": {
-        "php": "^7.1",
-        "laminas/laminas-eventmanager": "^2.6 || ^3.0",
-        "laminas/laminas-zendframework-bridge": "^1.0"
+        "php": "^7.3 || ~8.0.0",
+        "laminas/laminas-eventmanager": "^2.6 || ^3.3",
```
froschdesign: We can drop the support for laminas-eventmanager v2 here.

Slamdunk: Right, due to https://github.com/laminas/laminas-eventmanager/pull/10.

Done

composer.json
```diff
@@ -21,16 +21,16 @@
...
     },
     "require-dev": {
         "ext-phar": "*",
-        "doctrine/annotations": "^1.7",
-        "laminas/laminas-coding-standard": "^1.0",
-        "laminas/laminas-stdlib": "^2.7 || ^3.0",
-        "phpunit/phpunit": "^7.5.16 || ^8.4"
+        "doctrine/annotations": "^1.10.4",
+        "laminas/laminas-coding-standard": "^1.0.0",
+        "laminas/laminas-stdlib": "^2.7.7 || ^3.3.0",
```
froschdesign: If we support PHP 7.3 and higher then only version 3.3 of laminas-stdlib can be used.

weierophinney (approved):
:ship:

PR #56, opened 2020-11-30
#48 merge release `3.5.0` into development branch for `4.0.x`

Ocramius:
Fixes #48
Close #48

This is a full replacement for #48 which merges `3.5.0` into `4.0.x` (complex conflicts occurring there)


Ocramius:
Travis decided to bail on me. 

I'll add github actions as a workaround in a few, because I really need test visibility: otherwise navigating blind, and that's not a good scenario :-\

PR #54, opened 2020-11-29
#53 added support for generating parameters with union types in their definitions

fezfez:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

fix #53

PR #57, opened 2020-11-30
#53 support php-8.0 type system additions

### LINKED ISSUE #3: Not using return value of array_map in ClassGenerator->setImplementedInterfaces
https://github.com/zendframework/zend-code/blob/master/src/Generator/ClassGenerator.php#L476-L482 has the following snippet

```php
        array_map(function ($implementedInterface) {
            return (string) TypeGenerator::fromTypeString($implementedInterface);
        }, $implementedInterfaces);

        $this->implementedInterfaces = $implementedInterfaces;
```

That's a no-op, since the return value of the array_map isn't used. I think that the array_map should be removed, replaced with array_walk (probably pointless unless it would throw), or (probably) be changed to:

```php
        $this->implementedInterfaces = array_map(function ($implementedInterface) {
            return (string) TypeGenerator::fromTypeString($implementedInterface);
        }, $implementedInterfaces);
```

The latter seems likely given the rest of the changes in https://github.com/zendframework/zend-code/commit/ec09875ecc89347a6109e0eaea4a75e4d54dd8c3

This was detected via static analysis, I'm not sure of the impact of fixing this.

http://php.net/array_map

---

Originally posted by @TysonAndre at https://github.com/zendframework/zend-code/issues/165

### LINKED ISSUE #53: Internal class with parameters union type fail (php8)
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.5.0

#### Summary

Internal class with parameters union type fail (php8)

#### Current behavior
Exception with union type parameter

```
PHP Fatal error:  Uncaught Laminas\Code\Generator\Exception\InvalidArgumentException: Provided type "array|string|null" is invalid: must conform "/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*(\\[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*)*$/" in /var/www/src/Generator/TypeGenerator.php:74
Stack trace:
#0 /var/www/src/Generator/ParameterGenerator.php(185): Laminas\Code\Generator\TypeGenerator::fromTypeString('array|string|nu...')
#1 /var/www/src/Generator/ParameterGenerator.php(67): Laminas\Code\Generator\ParameterGenerator->setType('array|string|nu...')
#2 /var/www/test.php(8): Laminas\Code\Generator\ParameterGenerator::fromReflection(Object(Laminas\Code\Reflection\ParameterReflection))
#3 {main}
  thrown in /var/www/src/Generator/TypeGenerator.php on line 74
```

#### How to reproduce
```php
<?php

use Laminas\Code\Generator\ParameterGenerator;
use Laminas\Code\Reflection\ParameterReflection;

require 'vendor/autoload.php';

$parameter = ParameterGenerator::fromReflection(new ParameterReflection([\Phar::class, 'extractTo'], 1));
```

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

This is an overhaul of #53 by @fezfez which does:

 1. bump the minimum required PHP version to 7.4 (7.3 support ends this week)
 2. add extensive tests and improvements to `TypeGenerator` to accept various formats of PHP-8.0 type-related features (`mixed`, `false`, `null`, union)

Fixes #53
Fixes #54

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.3 || ~8.0.0",
+        "php": "^7.4 || ~8.0.0",
         "laminas/laminas-eventmanager": "^3.3"
     },
     "require-dev": {
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -109,10 +109,8 @@
       <code>generate</code>
       <code>removeTrait</code>
     </MissingReturnType>
-    <MixedArgument occurrences="15">
+    <MixedArgument occurrences="13">
       <code>$array['name']</code>
-      <code>$implemented</code>
-      <code>$implementedInterface</code>
       <code>$this-&gt;traitUsageGenerator-&gt;generate()</code>
       <code>$value</code>
       <code>$value</code>
@@ -126,7 +124,8 @@
       <code>array_values($method)</code>
       <code>array_values($property)</code>
     </MixedArgument>
-    <MixedArgumentTypeCoercion occurrences="1">
+    <MixedArgumentTypeCoercion occurrences="2">
+      <code>$interfaces</code>
       <code>$name</code>
     </MixedArgumentTypeCoercion>
     <MixedAssignment occurrences="7">
@@ -538,9 +537,6 @@
       <code>$parameter</code>
       <code>$value</code>
     </MixedAssignment>
-    <MixedOperand occurrences="1">
-      <code>$this-&gt;returnType-&gt;generate()</code>
-    </MixedOperand>
     <PossiblyFalseArgument occurrences="1">
       <code>$reflectionMethod-&gt;getDocBlock()</code>
     </PossiblyFalseArgument>
@@ -556,9 +552,6 @@
     <RedundantCastGivenDocblockType occurrences="1">
       <code>(bool) $returnsReference</code>
     </RedundantCastGivenDocblockType>
-    <RedundantCondition occurrences="1">
-      <code>method_exists($methodReflection, 'getReturnType')</code>
-    </RedundantCondition>
     <UnsafeInstantiation occurrences="2">
       <code>new static($array['name'])</code>
       <code>new static()</code>
@@ -587,14 +580,7 @@
     <MixedAssignment occurrences="1">
       <code>$value</code>
     </MixedAssignment>
-    <MixedOperand occurrences="1">
-      <code>$this-&gt;type-&gt;generate()</code>
-    </MixedOperand>
     <NullableReturnStatement occurrences="1"/>
-    <PossiblyNullReference occurrences="2">
-      <code>getName</code>
-      <code>getParentClass</code>
-    </PossiblyNullReference>
     <PropertyNotSetInConstructor occurrences="4">
       <code>$defaultValue</code>
       <code>$name</code>
@@ -607,10 +593,6 @@
       <code>(int) $position</code>
       <code>(string) $name</code>
     </RedundantCastGivenDocblockType>
-    <RedundantCondition occurrences="2">
-      <code>method_exists($reflectionParameter, 'getType')</code>
-      <code>method_exists($reflectionParameter, 'isVariadic')</code>
-    </RedundantCondition>
     <RedundantConditionGivenDocblockType occurrences="1">
       <code>$this-&gt;defaultValue instanceof ValueGenerator</code>
     </RedundantConditionGivenDocblockType>
@@ -797,27 +779,22 @@
     </MissingReturnType>
   </file>
   <file src="src/Generator/TypeGenerator.php">
-    <InvalidScalarArgument occurrences="4">
-      <code>$trimmedNullable</code>
-      <code>$trimmedType</code>
-      <code>$trimmedType</code>
-      <code>$trimmedType</code>
-    </InvalidScalarArgument>
-    <MissingReturnType occurrences="1">
-      <code>generate</code>
-    </MissingReturnType>
-    <MixedArgument occurrences="1">
-      <code>$this-&gt;generate()</code>
-    </MixedArgument>
-    <PossiblyInvalidPropertyAssignmentValue occurrences="2">
-      <code>$nullable</code>
-      <code>$trimmedType</code>
-    </PossiblyInvalidPropertyAssignmentValue>
-    <PropertyNotSetInConstructor occurrences="3">
-      <code>$isInternalPhpType</code>
-      <code>$nullable</code>
-      <code>$type</code>
-    </PropertyNotSetInConstructor>
+    <ImpureMethodCall occurrences="7">
+      <code>allowsNull</code>
+      <code>getName</code>
+      <code>getName</code>
+      <code>getName</code>
+      <code>getName</code>
+      <code>getParentClass</code>
+      <code>getTypes</code>
+    </ImpureMethodCall>
+    <MixedArgument occurrences="1"/>
+  </file>
+  <file src="src/Generator/TypeGenerator/AtomicType.php">
+    <LessSpecificReturnStatement occurrences="1"/>
+    <MoreSpecificReturnType occurrences="1">
+      <code>non-empty-string</code>
+    </MoreSpecificReturnType>
   </file>
   <file src="src/Generator/ValueGenerator.php">
     <DocblockTypeContradiction occurrences="2">
@@ -1338,9 +1315,13 @@
     </MissingReturnType>
   </file>
   <file src="test/Generator/ClassGeneratorTest.php">
-    <ArgumentTypeCoercion occurrences="2">
+    <ArgumentTypeCoercion occurrences="7">
+      <code>'Class1'</code>
       <code>'InvalidArgumentException'</code>
       <code>'LaminasTest_Code_NsTest_BarClass'</code>
+      <code>['Class1', 'Class2']</code>
+      <code>['Class1', 'Class2']</code>
+      <code>['Class1', 'Class2']</code>
     </ArgumentTypeCoercion>
     <InvalidArgument occurrences="5">
       <code>$resource</code>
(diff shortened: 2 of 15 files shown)


composer.json
```diff
@@ -20,7 +20,7 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^7.3 || ~8.0.0",
+        "php": "^7.4 || ~8.0.0",
```
Ocramius, line 23: Note: Travis-CI is dead anyway, so Github Actions **ABSOLUTELY MUST** be enabled before merging this patch.

Ocramius: `CHANGELOG.md` entries for the entire patch and this line change are needed.

weierophinney: I think that the note in the patch shouldn't be that we're bumping to 7.4 because 7.3 is unsupported (it is actually still supported; it's just getting security updates **only** for the coming year), but because _you are making use of PHP 7.4+ features_ (specifically, arrow functions and typed properties).

src/Generator/TypeGenerator.php
```diff
@@ -49,88 +51,160 @@ final class TypeGenerator implements GeneratorInterface
...
+                    $type
+                ));
+            }
+
+            foreach ($types as $index => $atomicType) {
+                $atomicType->assertCanUnionWith(array_diff_key($types, array_flip([$index])));
+            }
+        }
+
+        return new self($types, $nullable);
```
weierophinney:
```suggestion
        if (1 === count($types)) {
            $types[0]->assertCanBeAStandaloneType();

            if ($nullable) {
                $types[0]->assertCanBeStandaloneNullable();
            }
            return new self($types, $nullable);
        }
        
        if ($nullable) {
            throw new InvalidArgumentException(sprintf(
                'Type "%s" is a union type, and therefore cannot be also marked nullable with the "?" prefix',
                $type
            ));
        }
        
        foreach ($types as $index => $atomicType) {
            $atomicType->assertCanUnionWith(array_diff_key($types, array_flip([$index])));
        }

        return new self($types, $nullable);
```

I know my suggestion adds a redundant line, but it makes understanding the paths through the method a bit clearer.

src/Generator/TypeGenerator.php
```diff
@@ -49,88 +51,160 @@ final class TypeGenerator implements GeneratorInterface
...
     }
 
-    private function __construct()
+    /**
+     * @var AtomicType[]                     $types
+     *
```
weierophinney, line 160: Is it standard to have an empty line between standard and Psalm-specific annotations? (mainly curious, so I can be consistent!)

test/Generator/ParameterGeneratorTest.php
```diff
@@ -591,6 +590,46 @@ public function testGetInternalClassDefaultParameterValue()
...
+    public function php80Methods(): array
+    {
+        return [
+            [Php80Types::class, 'mixedType', 0, 'mixed', 'mixed $parameter'],
+            [Php80Types::class, 'falseType', 0, Php80Types::class . '|false', '\\' . Php80Types::class . '|false $parameter'],
+            [Php80Types::class, 'unionNullableType', 0, 'bool', '?bool $parameter'],
+            [Php80Types::class, 'unionReverseNullableType', 0, 'bool', '?bool $parameter'],
+            [Php80Types::class, 'unionNullableTypeWithDefaultValue', 0, 'bool|string|null', 'bool|string|null $parameter = null'],
+            [Php80Types::class, 'unionType', 0, Php80Types::class . '|' . stdClass::class, '\\' . Php80Types::class . '|\\' . stdClass::class . ' $parameter'],
+            [Php80Types::class, 'selfAndBoolType', 0, Php80Types::class . '|bool', '\\' . Php80Types::class . '|bool $parameter'],
```
weierophinney, line 629: Would be nice to have description keys for each of these; makes searching for specific cases on failure easier. The second string could very easily be re-purposed as a key for each of these.

test/Generator/TypeGeneratorTest.php
```diff
@@ -166,15 +162,48 @@ public function validType()
...
+            ['string|false', 'string|false'],
+            ['string|false|null', 'string|false|null'],
+
+            // `false` + `null` requires a third type
+            ['Foo|false|null', '\\Foo|false|null'],
+
+            // The `static` type should not be turned into a FQCN
+            ['static', 'static'],
+            ['?static', '?static'],
+            ['static|null', 'static|null'],
```
weierophinney, line 206: Same comment here. I know that the other cases don't have them, but no reason not to add them to new ones. It also makes them somewhat more self-describing:

```php
'false + null requires third type' => ['Foo|false|null', '\\Foo|false|null'],
```

weierophinney (changes requested):
Honestly, a nice strong set of changes. I have a few suggestions, but it's likely ready as-is.

composer.json
Ocramius: Works for me - I generally just need to get some speed on this package, and the various outdated approaches taken to types make it very fragile to work with, and reduce dev speed substantially :|

src/Generator/TypeGenerator.php
Ocramius: I got used to this from `doctrine/coding-standard`, where I group annotations by tool/rationale.

For example, `@param` annotations are split from `@return` by at least one empty line.

No convention though: if needed, can propose it for our own CS, but I'm not attached to this.

src/Generator/TypeGenerator.php
Ocramius: Removing one nesting level sounds fine - need to try it out and see how it looks :-)

test/Generator/TypeGeneratorTest.php
Ocramius: Unsure if I should add it though, because it obscures the default behavior of PHPUnit, which is to render the data provider values on failure :thinking:

test/Generator/ParameterGeneratorTest.php
Ocramius: Similar to https://github.com/laminas/laminas-code/pull/57#discussion_r532898401, adding keys makes it more readable here, but harder to understand on failure.

composer.json
```diff
@@ -20,7 +20,7 @@
...
         "laminas/laminas-eventmanager": "^3.3"
     },
     "require-dev": {
```
fezfez, line 26: I think you miss to add psalm in `require-dev` section

Ocramius: I did not introduce psalm at all, so far :D

fezfez: You have added some psalm annotation, maybe its a good start to introduce it to CI ?

Ocramius: Yes, that is part of #45 - I generally add psalm annotations anyway, as they provide much richer type information, and I have psalm running with the LSP plugin in my IDE :+1:

fezfez: Ok :+1:

fezfez (approved):
Sounds good to me, impressive work ! :+1:

Krienas:
Is this still missing something?

Ocramius:
Yes: time.

test/Generator/TypeGeneratorTest.php
Ocramius: Decided to skip on this: the types are indeed self-explanatory, and more grouped than singularly documented. Adding the keys is gonna lead to a regression in readability, in this specific case.

src/Generator/TypeGenerator.php
Ocramius: fc7d129f5d97cbb34f6b516a19ab5ac2d36f1d66

psalm-baseline.xml
```diff
@@ -126,7 +124,8 @@
       <code>array_values($method)</code>
       <code>array_values($property)</code>
     </MixedArgument>
-    <MixedArgumentTypeCoercion occurrences="1">
+    <MixedArgumentTypeCoercion occurrences="2">
+      <code>$interfaces</code>
```
Ocramius, line 128: The type for `$interfaces` was refined to `array<class-string>`, which obviously trips a lot of code locations where adding runtime checks now would be a major (and unnecessary) BC break.

psalm-baseline.xml
```diff
@@ -797,27 +779,22 @@
...
-      <code>$nullable</code>
-      <code>$trimmedType</code>
-    </PossiblyInvalidPropertyAssignmentValue>
-    <PropertyNotSetInConstructor occurrences="3">
-      <code>$isInternalPhpType</code>
-      <code>$nullable</code>
-      <code>$type</code>
-    </PropertyNotSetInConstructor>
+    <ImpureMethodCall occurrences="7">
+      <code>allowsNull</code>
```
Ocramius, line 783: Purity of `Reflection*#get*` APIs is not properly documented within psalm: potentially necessary to have upstream improvements, but nothing to fix here

psalm-baseline.xml
```diff
@@ -797,27 +779,22 @@
...
+      <code>getName</code>
+      <code>getParentClass</code>
+      <code>getTypes</code>
+    </ImpureMethodCall>
+    <MixedArgument occurrences="1"/>
+  </file>
+  <file src="src/Generator/TypeGenerator/AtomicType.php">
+    <LessSpecificReturnStatement occurrences="1"/>
+    <MoreSpecificReturnType occurrences="1">
+      <code>non-empty-string</code>
```
Ocramius, line 796: At this code location, the type inferred by psalm is `string` instead of `non-empty-string`, even though all code paths lead to non-empty strings being produced.

Potentially upstream issue to be fixed, but not a problem in our codebase.

psalm-baseline.xml
```diff
@@ -2857,6 +2849,19 @@
       <code>objectParameter</code>
     </MissingReturnType>
   </file>
+  <file src="test/TestAsset/Php80Types.php">
+    <ReservedWord occurrences="9">
```
Ocramius, line 2853: Not a problem: these reserved words are valid once we start using PHP 8, and will likely be improved in upstream

Ocramius:
I'll defer PHPCS upgrade to a new patch (will work on it now): the current CS ruleset does not understand PHP 7.4 symbols, and is causing a mess.

PR #58, opened 2020-11-30
BC BREAK: removed `Scanner` component, removed `FileReflection`, removed `NameInformation` and made `DocBlockScanner` `@internal`

### LINKED ISSUE #1: perfect FunctionScanner class
- [x] Add FunctionScanner for the new feature.
- [x] Add tests for the new feature.
- [x] Add a `CHANGELOG.md` entry for the new feature.

---

Originally posted by @pifeifei at https://github.com/zendframework/zend-code/pull/181

### LINKED ISSUE #5: [BC Break] Drop built-in code scanners for 4.0
There is a discussion in the PR #123 about possible removing of low-level scanners in the `Zend\Code\Scanner` subnamespace for 4.0. Reason for that is to avoid keeping support for low-level code analysis via tokens inside scanners, this job can be performed by external libraries, such as `nikic/php-parser`, `roave/better-reflection`, `goaop/parser-reflection`, etc.

Some facts about `Scanner` code usage in the project https://github.com/zendframework/zend-code/pull/123#issuecomment-382443064:

- There are only dependencies on `Scanner` namespace in the `Reflection` namespace, eg. `Zend\Code\Reflection\DocBlockReflection::reflect` uses `DocBlockScanner` class, as well `Zend\Code\Reflection\FileReflection::reflect` uses `CachingFileScanner`. All remaining integration parts could be removed as they unused.
- Classes from the `Reflection` namespace adds some public API to the PHP internal classes, thus `Reflection` namespace could not be easily dropped if we want to support `*Generator::fromReflection()` methods that accept only `Zend\Code\Reflection` classes and use specific methods from reflection.
- If we decide to drop `fromReflection()` method from all generators or replace typehint in them with built-in PHP classes, then whole `Scanner` and `Reflection` namespace could be dropped, but it will be significant BC break as we remove all functionality that are not related directly to the code generation.

@Ocramius suggestions from https://github.com/zendframework/zend-code/pull/123#issuecomment-382547482:

>> Whole `Scanner` namespace could be removed, there are only dependencies in `Reflection` namespace, eg. `Zend\Code\Reflection\DocBlockReflection::reflect` uses `DocBlockScanner` class, as well `Zend\Code\Reflection\FileReflection::reflect` uses `CachingFileScanner`. All remaining integration parts could be removed as they unused.

> Yeah, this stuff can be dropped. At this point, pointing to roave/better-reflection for most of the "brute-forcing the filesystem, looking for symbols"

>> Classes from the `Reflection` namespace adds some public API to the PHP internal classes, thus `Reflection` namespace could not be easily dropped if we want to support `*Generator::fromReflection()` methods that accept only `Zend\Code\Reflection` classes and use specific methods from reflection.

> Not sure about this one. I think the initial idea was that ext-reflection had huge potholes, and the library tried to fill them. I'd keep this for a separate issue, to figure out if the `Zend\Code\Reflection` namespace actually does anything more than ext-reflection these days. Considering that `Zend\Code\Reflection` is the entry point to all of the usages of this library (including the generators), I think that would make the migration path too painful at first.

>> If we decide to drop `fromReflection()` method from all generators or replace typehint in them with built-in PHP classes, then whole `Scanner` and `Reflection` namespace could be dropped, but it will be significant BC break as we remove all functionality that are not related directly to the code generation.

> The `::fromReflection()` constructors are extremely useful to mimick API without rewriting all of it manually, so I'd probably keep them in place. Instead, I think that allowing core reflection classes as parameters would be a better solution here.

>> I like the idea to drop both `Scanner` and `Reflection` namespaces from the code.

> Let's start with the `Scanner` only for now.

---

Originally posted by @lisachenko at https://github.com/zendframework/zend-code/issues/155

### LINKED ISSUE #8: Zend\Code\Reflection\FileReflection crashes on Zend's module.config.php
When trying to do a FileReflection on a zend module.config.php you run into a  ReflectionException when invoking FileReflection::getClasses().  Underlying problem is the tokens_get_all() function (in Zend\Code\Scanner\FileScanner::__construct) incorrectly flags the object::class constant as a T_CLASS.

---

Originally posted by @BavoJanss at https://github.com/zendframework/zend-code/issues/142

### LINKED ISSUE #9: Zend\Code\Reflection\FileReflection getBody()
When instantiating a Zend\Code\Generator\FileGenerator::fromReflection() one would expect ::generate to fully reconstruct any original file. Unfortunately Zend\Code\Reflection\FileReflection has no ::getBody method so any code outside classes is ignored.

---

Originally posted by @BavoJanss at https://github.com/zendframework/zend-code/issues/141

### LINKED ISSUE #12: Extra visual debt
This is just an attempt to fix the type-hinting mess inside the library. This is a BC break, as inheritance may be broken if the child classes do not respect the parent class signature.

Also, I suggest removing all the scanner stuff before merging.

---

Originally posted by @Ocramius at https://github.com/zendframework/zend-code/pull/123

### LINKED ISSUE #17: Deprecate scanner functionality (4.0), drop it (5.0) in favor of
Basically, parsing via regexes is simply not acceptable/viable anymore, and absolutely doesn't scale, as the syntax of PHP keeps changing.

@asgrim has been building https://github.com/Roave/BetterReflection since last year, and the library is stable/mature, so I'd suggest that we chop the head off this part of zend-code's functionality.

Thoughts?

---

Originally posted by @Ocramius at https://github.com/zendframework/zend-code/issues/91

### LINKED ISSUE #19: Support nullable types
As per PHP 7.1, it is possible to define nullability of a type by adding a `?` in front of it ( https://wiki.php.net/rfc/nullable_types ).

A few heads-up:
- careful about `?Foo $bar = null` vs `?Foo $bar` vs `Foo $bar = null`
- need to eventually patch the `TypeGenerator` to include the `?` in it (not sure yet)

---

Originally posted by @Ocramius at https://github.com/zendframework/zend-code/issues/85

### LINKED ISSUE #22: Update AnnotationScanner.php
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7113
User: @nouc23
Created On: 2015-01-09T10:38:17Z
Updated At: 2015-05-07T14:11:59Z
Body
I know we should write code using PSR standard but life is different.
While we use tabs script ignore annotations.

Fixes #5595

---
### Comment

User: @Ocramius
Created On: 2015-01-09T10:38:57Z
Updated At: 2015-01-09T10:38:57Z
Body
This requires a test asset containing tabs, and a related test

---
### Comment

User: @nouc23
Created On: 2015-01-09T10:58:53Z
Updated At: 2015-01-09T10:58:53Z
Body
done

---
### Comment

User: @weierophinney
Created On: 2015-02-19T20:50:49Z
Updated At: 2015-02-19T20:50:49Z
Body
Needs:
- Rebase
- new test.

---
### Comment

User: @Maks3w
Created On: 2015-05-07T13:57:11Z
Updated At: 2015-05-07T13:57:11Z
Body
I've pulled this and found the test is not related with the change made in scanner. So I don't know what is trying to fix this. Test never fails.

---
### Comment

User: @nouc23
Created On: 2015-05-07T14:11:59Z
Updated At: 2015-05-07T14:11:59Z
Body

I edit existing test, and I was told to create new one.
## I forgot to do that and also I doesnt use github much. I try to finish this up on days.

Wysłano z aplikacji myMail dla Androida
czwartek, 07 maja 2015, 03:57PM +02:00 od Maks3w notifications@github.com:

> I've pulled this and found the test is not related with the change made in scanner. So I don't know what is trying to fix this. Test never fails.
> —
> Reply to this email directly or  view it on GitHub .

---

---

Originally posted by @GeeH at https://github.com/zendframework/zend-code/issues/78

### LINKED ISSUE #27: Zend\Code\Scanner\MethodScanner does not account for scalar type declarations
I believe the issue is with `MethodScanner` and not `ParameterScanner`, as this is what is actually parsing the PHP tokens - it correctly ignores non-scalar type declarations such as `array` and `callable`. From what I can tell, `string`, `int`, `bool` and `float` are tokens of type `T_STRING`. As a result, the scanner believes they are classnames. This can result in classnames that simply don't exist.

I'm not really sure how this can be worked around inside the scanner. As far as I can tell, no PHP version checks are performed on the code being parsed, which means there's no reliable way of detecting whether or not `string` is an actual classname in PHP <7.0 code, or a scalar type declaration in PHP >=7.0 code.

Perhaps there needs to be some way to globally set the PHP version of the code being scanned?

---

Originally posted by @djmattyg007 at https://github.com/zendframework/zend-code/issues/56

### LINKED ISSUE #28: DirectoryScanner->getClasses() returns no classes
I scan with this class a directory for classes. But getClasses returns an empty array. When I test getFiles it works I get dozens files. I use php7.

---

Originally posted by @aight8 at https://github.com/zendframework/zend-code/issues/50

Ocramius:
The scanners were full of incomplete, untested and generally experimental code that would simply
crash when presented with something like PHP 8 code.

In addition to that, they were extremely unmaintainable due to the code being written for PHP 5,
and never being touched afterwards (too complex).

In order to slim down things and keep maintenance cost low, the complete `Laminas\Code\Scanner`
component has been trashed, with the exception of `Laminas\Code\Scanner\DocBlockScanner`, which
is still used in in some reflection internals.

It is advised for end-users to move to more modern tooling, such as [`roave/better-reflection`](https://github.com/Roave/BetterReflection)
which provides a safer (and well tested) API to scan source directories and extract reflection
information without loading any potentially dangerous PHP files.

The following components have been removed from the public API:

 * `Laminas\Code\Scanner\*`
 * `Laminas\Code\NameInformation` (unused in this package)
 * `Laminas\Code\Reflection\FileReflection` (use `roave/better-reflection` instead)
 * `Laminas\Code\Generic\Prototype\*`, which contains utilities to define custom docblock types:
   please consider using something like https://github.com/phpDocumentor/ReflectionDocBlock instead.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

TODOs:

 * [x] write `CHANGELOG.md` entry (992c26ec9f67496709dbb75d7b0af4aa11551043)
 * [x] check that documentation does not contain any scanner component references ( https://github.com/laminas/laminas-code/pull/58#pullrequestreview-541292257 )


Ocramius:
Also here: **BEFORE** merging, we need Github Actions to actually run on the codebase.

I went way overtime with this, but my aim is to do that Wednesday after work.

weierophinney (approved):
:ship:

I've checked the documentation and README.md file, and there are no mentions at all of scanning, so I think with a CHANGELOG.md entry, you're good to go.

PR #68, opened 2021-01-09
[Feature] Add/update type information for source files

lisachenko:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR adds/updates type information for source files to fix static analysis errors in dependent libraries. Types has been added to properties and phpDocs only to keep existing method signatures. So nothing should change at all.

### DIFF
--- a/src/Generator/ClassGenerator.php
+++ b/src/Generator/ClassGenerator.php
@@ -39,41 +39,38 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
     public const FLAG_ABSTRACT      = 0x01;
     public const FLAG_FINAL         = 0x02;
 
-    /** @var FileGenerator */
-    protected $containingFileGenerator;
+    protected ?FileGenerator $containingFileGenerator = null;
 
-    /** @var string */
-    protected $namespaceName;
+    protected ?string $namespaceName = null;
 
-    /** @var DocBlockGenerator */
-    protected $docBlock;
+    protected ?DocBlockGenerator $docBlock = null;
 
-    /** @var string */
-    protected $name;
+    protected string $name = '';
 
-    /** @var bool */
-    protected $flags = 0x00;
+    protected int $flags = 0x00;
 
-    /** @var string */
-    protected $extendedClass;
+    /** @psalm-var ?class-string */
+    protected ?string $extendedClass = null;
 
     /**
-     * @var string[] Array of string names
+     * Array of implemented interface names
+     *
+     * @var string[]
      * @psalm-var array<class-string>
      */
-    protected $implementedInterfaces = [];
+    protected array $implementedInterfaces = [];
 
     /** @var PropertyGenerator[] */
-    protected $properties = [];
+    protected array $properties = [];
 
     /** @var PropertyGenerator[] */
-    protected $constants = [];
+    protected array $constants = [];
 
     /** @var MethodGenerator[] */
-    protected $methods = [];
+    protected array $methods = [];
 
     /** @var TraitUsageGenerator Object to encapsulate trait usage logic */
-    protected $traitUsageGenerator;
+    protected TraitUsageGenerator $traitUsageGenerator;
 
     /**
      * Build a Code Generation Php Object from a Class Reflection
@@ -109,7 +106,6 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
 
         $interfaceNames = [];
         foreach ($interfaces as $interface) {
-            /** @var ClassReflection $interface */
             $interfaceNames[] = $interface->getName();
         }
 
@@ -139,7 +135,11 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
         $methods = [];
 
         foreach ($classReflection->getMethods() as $reflectionMethod) {
-            $className = $cg->getNamespaceName() ? $cg->getNamespaceName() . '\\' . $cg->getName() : $cg->getName();
+            $className     = $cg->getName();
+            $namespaceName = $cg->getNamespaceName();
+            if ($namespaceName !== null) {
+                $className = $namespaceName . '\\' . $className;
+            }
 
             if ($reflectionMethod->getDeclaringClass()->getName() == $className) {
                 $methods[] = MethodGenerator::fromReflection($reflectionMethod);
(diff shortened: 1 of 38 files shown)


src/Generator/ClassGenerator.php
```diff
@@ -39,41 +39,35 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
...
-    protected $flags = 0x00;
+    protected int $flags = 0x00;
 
-    /** @var string */
-    protected $extendedClass;
+    protected ?string $extendedClass = null;
 
     /**
      * @var string[] Array of string names
-     * @psalm-var array<class-string>
```
lisachenko, line 62: For codegen purposes we often use string names from different sources, thus static check for `class-string` will mostly fail, as it is impossible to add additional guards like `class_exists`, etc.

So, I'm suggesting to use parent `string` type here. WDYT?

src/Generator/ClassGenerator.php
```diff
@@ -139,7 +132,11 @@ public static function fromReflection(ClassReflection $classReflection)
         $methods = [];
 
         foreach ($classReflection->getMethods() as $reflectionMethod) {
-            $className = $cg->getNamespaceName() ? $cg->getNamespaceName() . '\\' . $cg->getName() : $cg->getName();
+            $className     = $cg->getName();
+            $namespaceName = $cg->getNamespaceName();
+            if ($namespaceName !== null) {
```
lisachenko, line 140: Check is required to avoid possible null concatenation warning in Psalm

src/Generator/ClassGenerator.php
```diff
@@ -211,14 +208,14 @@ public static function fromArray(array $array)
     }
 
     /**
-     * @param  string $name
-     * @param  string $namespaceName
-     * @param  array|string $flags
```
lisachenko, line 216: This docBlock was partially wrong, as it never accepts `string` as a value.

src/Generator/ClassGenerator.php
```diff
@@ -283,7 +280,7 @@ public function getName()
     }
 
     /**
-     * @param  string $namespaceName
+     * @param  ?string $namespaceName
```
lisachenko, line 287: There is an existing call in the code: `setNamespaceName(null)`, so param should be nullable.

src/Generator/ClassGenerator.php
Ocramius: IMO class-string to be kept: if symbol names are being invented, we declare them from the caller side, but in here we should work with strict guarantees.

lisachenko: @Ocramius I also want to have strict guarantees. But don't want to add suppress or custom docBlocks in client code. 

Consider this code from the`ClassGeneratorTest`:

```php
        $classGenerator->setImplementedInterfaces([
            'SomeNamespace\ClassInterface',
            GeneratorInterface::class,
            'Iteratable',
        ]);
```

How would you like to handle `class-string` guarantee for 1st and 3rd arguments? Psalm checks that only `::class` values can be passed and if we don't have an interface which we want to implement (eg. interface is being generated), then this check will always fail.

Possible solutions:
1) Add `@psalm-suppress' to such calls. Not a good idea..
2) Extract a variable and add a `@psalm-var array<class-string>` docBlock with proper type before calling `setImplementedInterfaces()`. Aka "visual debt". From client's POV this will be strange again. 

Is it ok for you to force usage of type casting if necessary to the client side to make sure that all interfaces are present?

Ocramius: > But don't want to add suppress or custom docBlocks in client code.

It's up to the client to ensure that the input is a valid class-alike symbol.

 > Aka "visual debt". From client's POV this will be strange again.

Not a problem either. We add types to push some validation constraints to the edge of the source (consumers), which is normal and happens every time types are being refined.

lisachenko: Ok, accepted 👍

test/Generator/ParameterGeneratorTest.php
```diff
@@ -133,8 +133,9 @@ public function testFromReflectionGetDefaultValueNotOptional()
         self::assertCount(2, $params);
 
         $firstParameter = ParameterGenerator::fromReflection($params[0]);
-        self::assertInstanceOf(ValueGenerator::class, $firstParameter->getDefaultValue());
-        self::assertNull($firstParameter->getDefaultValue()->getSourceContent());
+        $valueGenerator = $firstParameter->getDefaultValue();
+        self::assertInstanceOf(ValueGenerator::class, $valueGenerator);
+        self::assertSame('', $valueGenerator->getSourceContent());
```
Ocramius: This change looks suspicious

test/Generator/PropertyGeneratorTest.php
```diff
@@ -173,7 +173,9 @@ public function testPropertyWillLoadFromReflection(): void
         $cgProp = PropertyGenerator::fromReflection($reflProp);
 
         self::assertEquals('_bazProperty', $cgProp->getName());
-        self::assertEquals([true, false, true], $cgProp->getDefaultValue()->getValue());
+        $bazPropertyValue = $cgProp->getDefaultValue();
+        self::assertNotNull($bazPropertyValue);
+        self::assertEquals([true, false, true], $bazPropertyValue->getValue());
```
Ocramius: Should be `assertSame` to avoid type juggling issues

test/Generator/PropertyGeneratorTest.php
```diff
@@ -182,7 +184,9 @@ public function testPropertyWillLoadFromReflection(): void
         $cgProp = PropertyGenerator::fromReflection($reflProp);
 
         self::assertEquals('_bazStaticProperty', $cgProp->getName());
-        self::assertEquals(TestAsset\TestClassWithManyProperties::FOO, $cgProp->getDefaultValue()->getValue());
+        $bazStaticValue = $cgProp->getDefaultValue();
+        self::assertNotNull($bazStaticValue);
+        self::assertEquals(TestAsset\TestClassWithManyProperties::FOO, $bazStaticValue->getValue());
```
Ocramius: Should be `assertSame` to avoid type juggling issues

test/Generator/PropertyGeneratorTest.php
```diff
@@ -296,8 +300,10 @@ public function testSetDefaultValue(string $type, $value): void
         $property = new PropertyGenerator();
         $property->setDefaultValue($value, $type);
 
-        self::assertEquals($type, $property->getDefaultValue()->getType());
-        self::assertEquals($value, $property->getDefaultValue()->getValue());
+        $valueGenerator = $property->getDefaultValue();
+        self::assertNotNull($valueGenerator);
+        self::assertEquals($type, $valueGenerator->getType());
+        self::assertEquals($value, $valueGenerator->getValue());
```
Ocramius: Should be `assertSame` to avoid type juggling issues

Ocramius (approved):
Looking good: some minor issues (suspicious potential BC issues), and needs a rebase, but otherwise a really good improvement 👍

test/Generator/PropertyGeneratorTest.php
lisachenko: Done

test/Generator/ParameterGeneratorTest.php
lisachenko: Restored BC, now it uses null default value to keep BC.

lisachenko:
Hey, @Ocramius, could you please check this PR again?

My next PR with changes for probable next major version #69  will depend on this PR, so want to be sure that everything is ok with current PR.

Ocramius (approved):
LGTM :+1:

Ocramius:
Overall, the change in `protected` properties could be considered a BC break, but since the parent class depends on the type being respected, I'm OK with saying that subtypes have to respect it.

Ocramius:
Thanks @lisachenko!

Issue #35, opened 2020-01-15
`ParameterGenerator` always fails with variadic parameters


michalbundyra:
The `ParameterGenerator` always considers [`ReflectionParameter#isOptional()`](https://github.com/zendframework/zend-code/blob/3da502f243df7a189b9c1cfc70a96944ef111957/src/Generator/ParameterGenerator.php#L77) when checking for default parameter values.

A variadic argument is, though, always optional and never has a default value.

---

Originally posted by @Ocramius at https://github.com/zendframework/zend-code/issues/6

mchekin:
@michalbundyra 
I would like to take care of this issue.

mchekin:
@michalbundyra 

Maybe I do not understand the issue, but according to the source code:
https://github.com/laminas/laminas-code/blob/4.1.x/src/Generator/ParameterGenerator.php#L59-L65

And according to the tests:
https://github.com/laminas/laminas-code/blob/4.1.x/test/Generator/ParameterGeneratorTest.php#L506-L556

It seems like everything works as expected.

Am I missing something here?

Ocramius:
@mchekin the test checks `->isVariadic()`, but not `->isOptional()`, I think?

mchekin:
@Ocramius 
@michalbundyra 

I created a pull request which prevents generating variadic parameters with default value + an accompanying unit test https://github.com/laminas/laminas-code/pull/72

Ocramius:
Handled in #72

PR #72, opened 2021-01-30
Make sure variadic parameter does not have default value

mchekin:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description
Related to issue https://github.com/laminas/laminas-code/issues/35

### DIFF
--- a/test/Generator/ParameterGeneratorTest.php
+++ b/test/Generator/ParameterGeneratorTest.php
@@ -524,25 +524,8 @@ class ParameterGeneratorTest extends TestCase
         ));
 
         self::assertTrue($parameter->getVariadic());
-        self::assertNull($parameter->getDefaultValue());
+        self::assertTrue($parameter->isOptional());
         self::assertSame($expectedGeneratedSignature, $parameter->generate());
-    }
-
-    public function testGeneratingVariadicParameterWithDefaultValueThrowsInvalidArgumentException(): void
-    {
-        $parameter = new ParameterGenerator();
-
-        $parameter->setName('parameter');
-        $parameter->setType('int');
-        $parameter->setPosition(1);
-        $parameter->setVariadic(true);
-
-        $parameter->setDefaultValue([]);
-
-        $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessage('Variadic parameter cannot have a default value');
-
-        $parameter->generate();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


test/Generator/ParameterGeneratorTest.php
```diff
@@ -576,6 +577,23 @@ public function testSetGetVariadic()
...
+
+        $parameter->setName('parameter');
+        $parameter->setType('int');
+        $parameter->setPosition(1);
+        $parameter->setVariadic(true);
+
+        $parameter->setDefaultValue([]);
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Variadic parameter cannot have a default value');
```
Ocramius, line 606: I wonder if the exception should be raised by `->generate()` or by `->setDefaultValue()` and `->setVariadic()`? :thinking:

Ocramius (changes requested):

test/Generator/ParameterGeneratorTest.php
mchekin: @Ocramius 
I also considered this alternative.

The reason I went with `generate` is that: `generate()` is the central place that takes all the properties that were previously set and combines them to generate the output. It is always called and it is called last. 
So it is a good central place to validated if those properties make sense together, in my opinion.

If we put the validate in `->setDefaultValue()` and `->setVariadic()`, the two setters have to validate each other.
We would have to duplicate the check, because we do not know the order in which the two method would be called.
The validation logic is also would be distributed across the setters, instead of being in one central place.
And the two setters are now coupled to each other.
Thus we get more complexity.

What do you think?

Ocramius: > The validation logic is also would be distributed across the setters, instead of being in one central place.
 > And the two setters are now coupled to each other.
 > Thus we get more complexity.

Agree that it leads to more complexity, but it is still better than having a crash much later in the process.

To envision this more clearly:

```php
$parameterGenerator = someThirdPartyComponentCall();

// 10000 rows of code here

$parameterGenerator->generate(); // crash here - you will never understand what caused it
```

Therefore, I'd say that we need to rework this to include the exception in both setters: ugly, but necessary.

mchekin: @Ocramius 

Makes sense.
Thank you for the explanation.
Applied the changes.

Please check again.

mchekin:
@Ocramius 
Please review again.

Ocramius (approved):
LGTM, thanks @mchekin!

lisachenko:
Almost one year after )
This PR has introduced a BC break in minor version, related to the https://github.com/goaop/framework/issues/476

Ocramius:
@lisachenko test cases please :)

sakhunzai:
@lisachenko is right and I am unable to use the goaop framework due to this issue. Please fix it, thanks. The issues seems simple 

```
namespace Laminas\Code\Generator;
....
class ParameterGenerator extends AbstractGenerator
{ 
 /**
     * @param bool $variadic
     * @return ParameterGenerator
     */
    public function setVariadic($variadic)
    {
        if (isset($this->defaultValue)) {
            throw new Exception\InvalidArgumentException('Variadic parameter cannot have a default value');
        }

        $this->variadic = (bool) $variadic;

        return $this;
    }

```

The function should not throw exception, when $variadic is false. e. g

```
 public function setVariadic($variadic)
    {

       $this->variadic = (bool) $variadic;
        if ($this->variadic && isset($this->defaultValue)) {
            throw new Exception\InvalidArgumentException('Variadic parameter cannot have a default value');
        }

        return $this;
    }

```

Ocramius:
@sakhunzai send a patch with an attached test, please.

sakhunzai:
@Ocramius would you please review the PR https://github.com/laminas/laminas-code/pull/173 and merge, this is blocking my work, thanks

PR #73, opened 2021-02-18
Ensure scanner files php backwards compatibility in 3.5.x

hws47a:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

We are having an issue when updating to the 3.5.x on a project with PHP ^7.3.
I've prepared a fix in the same way it was done in 62efbb44f089cb46a548069e99f1b87e3593c52e

The PR with the issue:  https://github.com/magento/magento2/pull/32209

### DIFF
--- a/src/Scanner/ClassScanner.php
+++ b/src/Scanner/ClassScanner.php
@@ -930,6 +930,14 @@ class ClassScanner implements ScannerInterface
             return $infoIndex;
         };
 
+        // ensure php backwards compatibility
+        if (! defined('T_NAME_QUALIFIED')) {
+            define('T_NAME_QUALIFIED', 24001);
+        }
+        if (! defined('T_NAME_FULLY_QUALIFIED')) {
+            define('T_NAME_FULLY_QUALIFIED', 24002);
+        }
+
         /**
          * START FINITE STATE MACHINE FOR SCANNING TOKENS
          */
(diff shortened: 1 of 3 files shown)


hws47a:
Related to #47

hws47a:
Hi @weierophinney @Ocramius,

Could you please check and let me know if something like this can be merged to 3.5.x and potentially released in 3.5.2?
Alternatively we'll temporary define the constants in our application.

Regards,
Vladimir

Ocramius:
Hey @hws47a,  I would be OK to merge this, but I really really really need tests validating the change in order to ensure that this fix works on 3.x

hws47a:
Thanks Marco,

Could share how can I run Travis CI tests for this PR?

Ocramius:
Travis-CI is dead/defunct: what we.can do ia porting the new CI setup into
here.

For me, it would suffice to verify the tests locally first, then I can pick
up the CI part 👍

hws47a:
Short update: 

I found that in unit tests all classes are loaded using FileScanner class, this way you'll get missing constants defined before you call ClassScanner, MethodScanner, or ParameterScanner.

In our case, we were calling these classes directly and the errors were triggered.

So to replicate the issue we need tests that will call the classes directly. I'm also exploring the option to use FileScanner in our project.

hws47a:
It seems that if you use FileScanner to load the file before using other classes there is no issue.
We were able to handle the issue that way.

PR #75, opened 2021-04-14
Adding trait example into documentation.

### LINKED ISSUE #24: Examples for Trait generator
I had requested traits back in zendframework/zendframework#4876 which is now closed. It appears this support was added via TraitGenerator. Would be great to see some examples included in the docs.

---

Originally posted by @RSully at https://github.com/zendframework/zend-code/issues/73

vrkansagara:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
- adding example for the issue #24

### DIFF
--- a/docs/book/generator/examples.md
+++ b/docs/book/generator/examples.md
@@ -370,3 +370,46 @@ $generator->addMethod(
 );
 $code = $generator->generate();
 ```
+
+
+## Generating Trait 
+
+`Laminas\Code\Generator\FileGenerator` can be used to generate the contents of a *PHP* file. You can
+include classes as well as arbitrary content body.
+
+```php
+use Laminas\Code\Generator\FileGenerator;
+use Laminas\Code\Generator\TraitGenerator;
+
+$traitclass = new TraitGenerator();
+$traitclass->addMethod('MethodOne');
+$traitclass->addConstant('VERSION','1.0.0');
+
+$traitFile = new Laminas\Code\Generator\FileGenerator();
+$traitFile->setClass($traitclass);
+
+```
+
+Calling `generate()` will generate the code -- but not write it to a file. You will need to capture
+the contents and write them to a file yourself. As an example:
+
+```php
+$code = $traitFile->generate();
+file_put_contents('Trait.php', $traitclass);
+```
+
+The above will generate the following file:
+
+```php
+<?php
+
+trait 
+{
+    public const VERSION = '1.0.0';
+
+    public function MethodOne()
+    {
+    }
+}
+
+```


vrkansagara:
Ping @michalbundyra

Ocramius:
4091991 should probably not be part of this patch: was the patch started from this branch?

vrkansagara:
yes @Ocramius  , I think there is some webhook executed by @michalbundyra  which add 4091991

Ocramius:
@vrkansagara nay, that commit is in `master` and `3.4.x` since a while - I think the edit was simply done starting from the wrong branch. No big deal, but requires manual merge, rather than merge from here.

docs/book/generator/examples.md
```diff
@@ -370,3 +370,46 @@ $generator->addMethod(
 );
 $code = $generator->generate();
 ```
+
+
+## Generating Trait 
+
+`Laminas\Code\Generator\FileGenerator` can be used to generate the contents of a *PHP* file. You can
+include classes as well as arbitrary content body.
```
froschdesign, line 378: These two sentences are copied from the [section "Generating PHP files"](https://docs.laminas.dev/laminas-code/generator/examples/#generating-php-files) and it does not the describe the generating of a trait.

docs/book/generator/examples.md
```diff
@@ -370,3 +370,46 @@ $generator->addMethod(
...
+use Laminas\Code\Generator\TraitGenerator;
+
+$traitclass = new TraitGenerator();
+$traitclass->addMethod('MethodOne');
+$traitclass->addConstant('VERSION','1.0.0');
+
+$traitFile = new Laminas\Code\Generator\FileGenerator();
+$traitFile->setClass($traitclass);
+
+```
```
froschdesign, line 391: The code example can be shortened to the essentials.
```suggestion
```php
$trait = new Laminas\Code\Generator\TraitGenerator();
$trait->setName('TranslatorAwareTrait');
$trait->addProperty(
    'translator', // Name
    null, // Default value
    Laminas\Code\Generator\PropertyGenerator::FLAG_PRIVATE // Flags
);
$trait->addMethod(
    'getTranslator', // Name
    [], // Parameters
    Laminas\Code\Generator\PropertyGenerator::FLAG_PUBLIC, // Flags
    'return $this->translator;' // Body
);

echo $trait->generate();
```
```

File: docs/book/generator/examples.md [hunk 3]
```diff
@@ -370,3 +370,46 @@ $generator->addMethod(
...
+trait 
+{
+    public const VERSION = '1.0.0';
+
+    public function MethodOne()
+    {
+    }
+}
+
+```
```
froschdesign, line 415:
```suggestion
The code generator produces the following trait:

```php
trait TranslatorAwareTrait
{
    private $translator = null;

    public function getTranslator()
    {
        return $this->translator;
    }
}
```
```

**Review by froschdesign — CHANGES_REQUESTED** [review]
Unfortunately, too much was copied from a previous section. Some corrections are needed.

**Comment by RSully** [comment]
Is it possible to add code to the body of a generated trait method? That would seem important and is missing from these examples, and part of why I made my initial request.

**Comment by vrkansagara** [comment]
> Is it possible to add code to the body of a generated trait method? That would seem important and is missing from these examples, and part of why I made my initial request.

@RSully   I have already explained here https://github.com/laminas/laminas-code/pull/75/commits/847be56c313cc9466b3a5198915643dad7db6304#diff-3d8f5388de37ae1d660bf9198127eb4e4a7d0796e5cd53fab4d602293b388371R406

**Comment by RSully** [comment]
That doesn't address my question. For example, I am asking to show something like this:

```php
trait HelloWorld {
    private $greeting = 'Hello World';

    public function sayHello() {
        echo $this->greeting;
    }
}
```

Also, according to a quick test in PHP 8, traits cannot have constants - so your example's generated code doesn't compile.

> Fatal error: Traits cannot have constants in [...]

froschdesign:
@RSully 
> Is it possible to add code to the body of a generated trait method?

I have updated the example from above.

docs/book/generator/examples.md
RSully: I think the `use` statements could still be handy - though I admit I don't know how the rest of the documentation is structured and if there's a standard or not.

Edit: never-mind ignore me :) - I see the classes are fully qualified

Ocramius:
Houskeeping is here, closing :)

PR #77, opened 2021-04-23
Consider missing indices in PropertyGenerator

### LINKED ISSUE #76: PropertyGenerator::fromReflection throws E_NOTICE for non-default typed properties
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 4.2.0

Apparently, `\ReflectionClass::getDefaultProperties` doesn't return typed properties without a default value, thus an `E_NOTICE` is emitted when assigning [`$defaultValue`](https://github.com/laminas/laminas-code/blob/4.3.x/src/Generator/PropertyGenerator.php#L38) because it's trying to access an undefined index.

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fix missing array index `E_NOTICE` error in `PropertyGenerator`.

fixes #76

Edit: The test explicitly checks, that typed properties will be ignored when generating the code. Maybe that feature should be implemented in a future release, and that will cause this test to fail.

### DIFF
--- a/test/Generator/TestAsset/ClassWithTypedProperty.php
+++ b/test/Generator/TestAsset/ClassWithTypedProperty.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\Code\Generator\TestAsset;
+
+final class ClassWithTypedProperty
+{
+    private string $typedProperty;
+
+    public function __construct(string $typedProperty)
+    {
+        $this->typedProperty = $typedProperty;
+    }
+}
--- a/src/Generator/PropertyGenerator.php
+++ b/src/Generator/PropertyGenerator.php
@@ -35,7 +35,7 @@ class PropertyGenerator extends AbstractMemberGenerator
 
         $allDefaultProperties = $reflectionProperty->getDeclaringClass()->getDefaultProperties();
 
-        $defaultValue = $allDefaultProperties[$reflectionProperty->getName()];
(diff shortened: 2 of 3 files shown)


Ocramius (approved):
Thanks @rieschl!

test/Generator/TestAsset/ClassWithTypedProperty.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\Code\Generator\TestAsset;
+
+final class ClassWithTypedProperty
+{
+    public string $typedProperty;
```
Ocramius: While annoying, it is true that this needs a constructor declared: could you please add one?

Ocramius (changes requested):

test/Generator/TestAsset/ClassWithTypedProperty.php
rieschl: thanks! sure, I just pushed a fix.

rieschl:
This bug is also in the 3.x branch (checked with 3.4.1). Do I have to submit a patch for this version, too?

Ocramius:
> This bug is also in the 3.x branch (checked with 3.4.1). Do I have to submit a patch for this version, too?

IMO no need, unless it it a security issue (not the case - codegen is not supposed to be built upon user input)

Ocramius:
Thanks @rieschl!

PR #81, opened 2021-05-06
Fix `DocBlockGenerator` formatting issue when docblock long description is not available

### LINKED ISSUE #80: DocBlockGenerator output breaks when used without $longDescription
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 4.2.1

#### Summary

`DocBlockGenerator` output formatting breaks when used without "long description".

#### Current behavior

When omitting the `$longDescription` argument, the constructur defaults to an empty string. Later checks (that assert if a `longDescription` is set) however test for `=== null` and never catch on to the fact that no description was supplied at all.

#### How to reproduce

Instantiate a `DocBlockGenerator` with empty `$longDescription` argument (2nd constructor argument):

```php
$dbg = new \Laminas\Code\Generator\DocBlockGenerator(
    "Foo",
    null,
    [new GenericTag("var", "array")]
);

echo $dbg->generate();
```

Actual output:

```
/**
 * Foo
 *
 *
 *
 * @var array
 */
```

#### Expected behavior

`$longDescription` should be able to be omitted (`=== null`) without generating comments with needlessly many LFs.

Expected output:

```
/**
 * Foo
 *
 * @var array
 */
```

#### Suggested solution

Do not initialize `$longDescription` (and probably also `$shortDescription`) with empty strings, but with `null` instead.

Alternatively, do not test for `!== null` when asserting that a long (or short) description is set, but just test for being falsy.

martin-helmich:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fixes #80.

See there for issue description.

In #80 I suggested two possible solutions:

> Do not initialize $longDescription (and probably also $shortDescription) with empty strings, but with null instead.
>
>Alternatively, do not test for !== null when asserting that a long (or short) description is set, but just test for being falsy.

To initialize `$longDescription` with `null`, the type hint of that property would need to be changed from `string` to `?string`, which (since the property is `protected`) might be considered a breaking change.

For that reason, I opted to implement my alternative solution, which effectively results in treating an empty (but non-null) "long description" as equivalent to "not set".

### DIFF
--- a/src/Generator/DocBlockGenerator.php
+++ b/src/Generator/DocBlockGenerator.php
@@ -228,10 +228,10 @@ class DocBlockGenerator extends AbstractGenerator
         }
 
         $output = '';
-        if (null !== ($sd = $this->getShortDescription())) {
+        if ($sd = $this->getShortDescription()) {
             $output .= $sd . self::LINE_FEED . self::LINE_FEED;
         }
-        if (null !== ($ld = $this->getLongDescription())) {
+        if ($ld = $this->getLongDescription()) {
             $output .= $ld . self::LINE_FEED . self::LINE_FEED;
         }
 
--- a/test/Generator/DocBlockGeneratorTest.php
+++ b/test/Generator/DocBlockGeneratorTest.php
@@ -202,4 +202,21 @@ EOS;
         $tags = $this->reflectionDocBlockGenerator->getTags();
         self::assertInstanceOf(ReturnTag::class, $tags[3]);
     }
+
+    public function testGenerateOmitsLongDescriptionWithTags(): void
+    {
+        $generator = new DocBlockGenerator(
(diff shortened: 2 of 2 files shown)


Ocramius:
Heyo, could you please retarget this to either 4.2.x or 4.3.x?

martin-helmich:
@Ocramius I targeted against master because the PR template told me to. I now rebased onto 4.2, and retarged the PR! :+1:

martin-helmich:
:new: Amended my change to fix coding style issues.

Ocramius (approved):
Excellent, thanks @martin-helmich!

PR #85, opened 2021-05-17
Normalize composer

ghostwriter:
**[QA]**
* Normalize `composer.json` and/or update `composer.lock`
    - Restructure `composer.json` according to the underlying [JSON schema](https://getcomposer.org/schema.json).
    - Prevent inconsistencies between all (200+) repositories across all 3 GitHub organizations (`laminas`, `mezzio`, and `laminas-api-tools`).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,27 +1,22 @@
 {
     "name": "laminas/laminas-code",
     "description": "Extensions to the PHP Reflection API, static code scanning, and code generation",
-    "license": "BSD-3-Clause",
     "keywords": [
         "laminas",
         "laminasframework",
         "code"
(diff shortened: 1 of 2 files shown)


Ocramius:
Looks like `composer.lock` was updated without considering the fact that we need to operate under the constraints of our CI setup: need to close this therefore.

Please consider re-opening if you want to run a `composer update nothing` instead.

ghostwriter:
okay, re-open it. I'll correct the PR.

Ocramius:
Excellent, thanks!

Ocramius (approved):

PR #86, opened 2021-06-05
Support for PHP 8.1

derrabus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

PHP 8.1 will introduce a process to add return types to internal methods, see https://wiki.php.net/rfc/internal_method_return_types

This library extends various reflection classes and overrides public methods that are affected by this change. PHP 8.1 triggers deprecation warnings unless we declare that a future version of this library will add these return types as well. We can do so by adding the `ReturnTypeWillChange` attribute to those methods.

This PR suggests to do just that because adding the actual return types would break code that extends our classes. A potential version 5 of this library could then add the actual return types.

There is one problematic case though: `MethodReflection::getPrototype()`. That method as it is currently implemented will return an array or string, depending on a parameter passed to it. The parent method `ReflectionMethod::getPrototype()` however will return a `ReflectionMethod` instance.

My recommended course of action would be to introduce two new methods `getPrototypeAsString()` and `getPrototypeAsArray()` and deprecate calling `MethodReflection::getPrototype()` with the `PROTOTYPE_AS_ARRAY`/`PROTOTYPE_AS_STRING` parameter.

If you agree with that, I could work on a follow-up PR.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -19,6 +19,7 @@ jobs:
         php-version:
           - "7.4"
           - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
--- a/test/Generator/ParameterGeneratorTest.php
+++ b/test/Generator/ParameterGeneratorTest.php
@@ -118,6 +118,7 @@ class ParameterGeneratorTest extends TestCase
 
     /**
      * @group 95
+     * @requires PHP < 8.1
      */
     public function testFromReflectionGetDefaultValueNotOptional()
     {
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "^7.4 || ~8.0.0"
+        "php": ">=7.4, <8.2"
     },
     "conflict": {
         "phpspec/prophecy": "<1.9.0"
@@ -19,7 +19,7 @@
         "doctrine/annotations": "^1.10.4",
         "laminas/laminas-coding-standard": "^2.3.0",
         "laminas/laminas-stdlib": "^3.3.0",
-        "phpunit/phpunit": "^9.4.2",
+        "phpunit/phpunit": "^9.5.9",
         "psalm/plugin-phpunit": "^0.14.0",
         "vimeo/psalm": "^4.3.1"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "84d0d82a51f336472dea1684ccd19451",
+    "content-hash": "9f7eb64d978b595fb90e1b7faf8ed85f",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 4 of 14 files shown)


derrabus:
Thanks for approving the CI run. Two static analysis jobs are failing:

* The PHP CodeSniffer version used here apparently does not know how to handle attributes and complains about "Perl-style comments" and an unused import `ReturnTypeWillChange`.
* Psalm complains about an undefined class `ReturnTypeWillChange`. We could either suppress that error or add `symfony/polyfill-php81` to `require-dev` to make the class known.

Please advice how I may proceed.

Ocramius:
ReturnTypeWillChange could either be stubbed, or we can wait to run psalm
on 8.1 (after this summer, when the first RC are verifiable)

derrabus:
`symfony/polyfill-php81` contains a stub, so I've added that package.

Ocramius:
No, stubbed as in stubbed for psalm, not further deps (and complexity) for
runtime users.

A polyfill is absolutely to be avoided here.

Ocramius:
Nvm, I see it is require-dev: will probably still need to be extracted to a
standalone file 🤔

Ocramius:
Marking as blocked until we test also on `8.1.x` (later this year)

derrabus:
Sure, whenever you're ready.

derrabus:
I had to bump PHP for the PHP CodeSniffer job to 8.0 because it apparently does not understand attributes when run on PHP 7.4. 😢

.github/workflows/phpunit.yml
```diff
@@ -43,6 +44,12 @@ jobs:
           key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
           restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
 
+      - name: "Fake PHP version"
+        if: ${{ matrix.php-version == '8.1' }}
+        run: |
+          composer config platform.php 8.0.99
+          composer update --lock
```
derrabus: This hack can be removed as soon as all of our dependencies declare compatibility with PHP 8.1.

test/Generator/ParameterGeneratorTest.php
```diff
@@ -118,6 +118,7 @@ public function testFromReflectionGetDefaultValue()
 
     /**
      * @group 95
+     * @requires PHP < 8.1
```
derrabus, line 121: I'm skipping the test on PHP 8.1 because PHP has changed its behavior here.

See https://3v4l.org/BocvA/rfc#vgit.master Vs. https://3v4l.org/BocvA

derrabus (commented):
Hello again. I've updated the PR to include a green CI for PHP 8.1. Please let me know if I can do anything else to help.

derrabus:
Hello again. This PR is one of the last blockers for Symfony's PHP 8.1 CI. Before I implement workarounds into our CI to make it pass despite the deprecation warnings triggered by laminas-code, I'd like to ask if there's anything I can do to move this PR forward.

Ocramius:
We'll wait for the next PHP RC before moving forward.

Before PHP 8.1 compatibility is released, we need to add intersection types
too.

derrabus:
All right. Thanks for the heads-up.

.github/workflows/phpunit.yml
alexander-schranz: Prophecy seems to be released and so this should not longer be needed :)

alexander-schranz (commented):
Tested [sulu](https://github.com/sulu/sulu/pull/6208) with symfony 5.3 against this PR and did work. Thx for all working on this 👍

.github/workflows/phpunit.yml
derrabus: `laminas/laminas-coding-standard` would need to be bumped as well. But that should probably be done in a separate PR.

derrabus: #98

Ocramius: #98 merged - I think we need to upgrade some of the CI tooling, but this one can probably be merged as-is after a rebase

.github/workflows/coding-standards.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "7.4"
```
Ocramius: We should probably run CS checks on 7.4, unless there's a reason to run them on newer versions (some CS checks use the PHP version in the runtime, sometimes)

derrabus: There was a reason I had to bump back then, but maybe this has been resolved in the meantime. I'll have a look.

.github/workflows/phpunit.yml
derrabus: To remove this hack, #98 was actually enough. I've removed it after the rebase.

.github/workflows/coding-standards.yml
derrabus: Reverted

Ocramius (approved):
🚢 

Thanks @derrabus!

derrabus:
Thank you for the review & merge, @Ocramius!

racile:
https://github.com/symfony/symfony/issues/44285

Ocramius:
@racile see the related milestone: https://github.com/laminas/laminas-code/milestone/19

Likely done around end of year, I'd say, but no promises.

PR #87, opened 2021-06-05
Pass `0` instead of `null` to `preg_match()` `$flags`, since in PHP 8.1+ the signature may change

derrabus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

The `$flags` parameter of `preg_match()` should be an integer, but `DocBlockScanner` passes `null` when calling the method. PHP 8.1 will be stricter about that and trigger a deprecation warning. This PR proposes to pass `0` instead, which is the actual default value of that parameter.

### DIFF
--- a/src/Scanner/DocBlockScanner.php
+++ b/src/Scanner/DocBlockScanner.php
@@ -176,7 +176,7 @@ class DocBlockScanner
             }
             $currentChar = $stream[$streamIndex];
             $matches     = [];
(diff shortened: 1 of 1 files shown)


Ocramius:
LGTM, but beware that this library does not allow installation with 8.1
yet, so that will be a question for ~ october, when the first RC releases
can be verified

Ocramius (approved):

Ocramius:
Thanks @derrabus!

PR #88, opened 2021-06-17
Drop unused `laminas/laminas-eventmanager` dependency, remove replacement of `zendframework/zend-code`

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Since `laminas/laminas-code:4.0.0`, we no longer have code that relies on `laminas/laminas-eventmanager`.
    
The dependency is still dragged in though (oversight), and that leads to indirectly installing `laminas/laminas-zendframework-bridge`, which has a runtime performance impact on the entire autoloading stack (see https://github.com/laminas/laminas-zendframework-bridge/issues/83 ).

Hereby we safely drop the dependency.

Since `zendframework/zend-code:4.0.0` and newer do not exist, newer versions of this library don't really provide a replacement for it anymore, and the replacement clause can be safely dropped.
    
This should lead to some downstream dependency graph simplifications.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,6 @@
     "require": {
         "php": "^7.4 || ~8.0.0"
     },
-    "replace": {
-        "zendframework/zend-code": "self.version"
-    },
     "conflict": {
         "phpspec/prophecy": "<1.9.0"
     },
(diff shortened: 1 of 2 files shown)


Xerkus (approved):

Ocramius:
:ship:

Issue #69, opened 2021-01-10
[RFC]: Next major version of package. Stricter types for methods, API cleaning.


lisachenko:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 5.x.x  |
| BC Break?           | **Yes** |

### Goal

I want to discuss next major version of this package and what can be improved in it to have better API for this package.

### Background

Current version of package contains too many places where method of a class can accept different types at the same time:  string/arrays/specific objects. This makes code error-prone and harder to debug, harder to use and increases chances of errors as IDE doesn't help too much when all checks are in runtime. Also static analysis tools complain about types for this library.

### Proposal(s)

1) To add `declare(strict_types=1)` to all source files to benefit more from stronger type checks
2) To to add type information to all method signatures (parameter and return types), which will be obviously a BC break, but it will improve code quality a lot.
3) As 2) will bring a BC break, I want to propose to drop all `public static function fromArray(array $array)` static constructors from the code, as all these constructors accept non-typed arrays, thus IDE doesn't know about possible keys and can't verify types with static analysis tools. Same should be applied to the `AbstractGenerator::setOptions($options)` - IMO, it should be removed for the next major version.
4) Drop all optional arguments from constructors, making ctor signatures smaller and cleaner, eg instead of 
```
    ParameterGenerator::__construct(
        $name = null,
        $type = null,
        $defaultValue = null,
        $position = null,
        $passByReference = false
    );
``` 
new signature will be 
```
    ParameterGenerator::__construct(string $name);
```
As we already have mutation methods in the API, we can use fluent interfaces to adjust generated code item by making required changes:
```
$parameterGenerator = new ParameterGenerator('test');
$parameterGenerator
    ->setType(TypeGenerator::fromTypeString('string'))
    ->setDefaultValue('foo')
    ->setPassByReference(true);
```
This should make interface for configuration more stable, as there will be only one way to configure generator via setters that can mutate instance (no more `fromArray`, no more tons of args in ctors).

5) Drop all mixed types for method signatures, eg:
```
    /**
     * Set the default value of the parameter.
     *
     * Certain variables are difficult to express
     *
     * @param  null|bool|string|int|float|array|ValueGenerator $defaultValue
     * @return ParameterGenerator
     */
    public function setDefaultValue($defaultValue);
```
will be replaced with following signature:
```
    /**
     * Set the default value of the parameter.
     */
[...]

PR #90, opened 2021-07-07
Fix `MethodGenerator` parameter sorting, when an explicit sorting is provided

willjones9:
The MethodGenerator doesn't seem to take any notice of the ParameterGenerator position property. This causes issues when adding properties out of order, especially with optional and variadic parameters.

### DIFF
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -15,6 +15,7 @@ use function strlen;
 use function strtolower;
 use function substr;
 use function trim;
+use function usort;
 
 class MethodGenerator extends AbstractMemberGenerator
 {
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -318,6 +318,7 @@ class MethodGenerator extends AbstractMemberGenerator
 
     /**
      * Sort parameters by their position
+     * @return void
      */
     protected function sortParameters()
     {
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -312,10 +312,8 @@ class MethodGenerator extends AbstractMemberGenerator
 
     /**
      * Sort parameters by their position
-     *
-     * @return void
      */
-    protected function sortParameters()
+    private function sortParameters(): void
     {
         usort($this->parameters, static function (ParameterGenerator $item1, ParameterGenerator $item2) {
             return $item1->getPosition() <=> $item2->getPosition();
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -30,8 +30,6 @@ class MethodGenerator extends AbstractMemberGenerator
 
     private bool $returnsReference = false;
 
-    private bool $sortParametersOnSet = true;
-
     /**
      * @return MethodGenerator
      */
@@ -215,13 +213,10 @@ class MethodGenerator extends AbstractMemberGenerator
      */
     public function setParameters(array $parameters)
     {
-        $this->sortParametersOnSet = false;
-
         foreach ($parameters as $parameter) {
             $this->setParameter($parameter);
         }
 
-        $this->sortParametersOnSet = true;
         $this->sortParameters();
 
         return $this;
@@ -252,9 +247,8 @@ class MethodGenerator extends AbstractMemberGenerator
 
         $this->parameters[$parameter->getName()] = $parameter;
 
-        if ($this->sortParametersOnSet) {
-            $this->sortParameters();
-        }
+        $this->sortParameters();
+
         return $this;
     }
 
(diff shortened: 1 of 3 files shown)


src/Generator/MethodGenerator.php
```diff
@@ -29,6 +30,8 @@ class MethodGenerator extends AbstractMemberGenerator
 
     private bool $returnsReference = false;
 
+    private bool $sortParametersOnSet = true;
```
Ocramius: Why is this required?

Ocramius (changes requested):

src/Generator/MethodGenerator.php
willjones9: `setParameters` calls `setParameter` in a loop and `setParameter` calls `sortParameters`. Having `sortParameters` called multiple times seems a little inefficient, so `$sortParametersOnSet` gets set to `false` at the start of `setParameters` to avoid needless calls.

Ocramius: @willjones-stratagem consider that `laminas/laminas-code` is not designed for runtime usage anyway: only for building sources.

I suggest removing any state mutation, and re-performing the operation, even if inefficient at first :-)

willjones9: @Ocramius fair shout. I'll update that.
That said, the project I'm working on at the moment will probably end up using `laminas/laminas-code` as part of certain runtime functionality so some overall efficiency savings in [5.x.x](https://github.com/laminas/laminas-code/issues/69) wouldn't be a bad thing

Ocramius: That is alright, but I'd like to prevent state mutation as much as possible.

An exception thrown during a flag switch would cause a mess, for example, and subclassing makes this much more complex too.

An alternate design will be needed, if this performance impact is relevant in future.

willjones9: Very good point. The PR's now been updated.

src/Generator/MethodGenerator.php
```diff
@@ -305,6 +310,18 @@ public function setReturnsReference($returnsReference)
         return $this;
     }
 
+    /**
+     * Sort parameters by their position
+     *
+     * @return void
+     */
+    protected function sortParameters()
```
Ocramius: We can declare `: void` here.

Method can also be `private`.

test/Generator/MethodGeneratorTest.php
```diff
@@ -74,6 +74,38 @@ public function testMethodParameterMutator()
         $methodGenerator->setParameter(new stdClass());
     }
 
+    public function testMethodParameterSetPosition()
```
Ocramius: We should have two separate test methods:

 * one verifying `setParameter()`
 * one verifying `setParameters()`

test/Generator/MethodGeneratorTest.php
```diff
@@ -74,6 +74,38 @@ public function testMethodParameterMutator()
...
+            ParameterGenerator::fromArray(['name' => 'baz', 'type' => stdClass::class, 'position' => 1]),
+            $bar
+        );
+
+        /** @var ParameterGenerator $baz */
+        $baz = array_shift($params);
+        self::assertSame('bar', $baz->getName());
+
+        $this->expectException(InvalidArgumentException::class);
+        $methodGenerator->setParameter(new stdClass());
```
Ocramius: This shouldn't be needed: it's not valid from a type perspective, and does not need a runtime test.

test/Generator/MethodGeneratorTest.php
```diff
@@ -74,6 +74,38 @@ public function testMethodParameterMutator()
...
+        $methodGenerator->setParameter(['name' => 'bar', 'type' => 'array', 'position' => 2]);
+        $methodGenerator->setParameter(
+            ParameterGenerator::fromArray(['name' => 'baz', 'type' => stdClass::class, 'position' => 1])
+        );
+
+        $params = $methodGenerator->getParameters();
+        self::assertCount(3, $params);
+
+        /** @var ParameterGenerator $foo */
+        $foo = array_shift($params);
```
Ocramius, line 88: Suggestion:

 1. `$sorting = array_map(static function (ParameterGenerator $parameter): string { return $parameter->getName(); }, $params);`
 2. `self::assertEquals(['foo', 'bar', 'baz'], $sorting);`

Would make the test much simpler to read, IMO, and removes the need for counting returned values.

Ocramius (changes requested):

Ocramius:
@willjones-stratagem I just noticed that this patch targets `4.5.x`, but this is a bugfix.

Should we change the target branch to `4.4.x`?

willjones9:
@Ocramius retargeting it works for me

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @willjones-stratagem!

PR #95, opened 2021-07-29
Fix incorrect argument type when `@param` is not declared for all parameters of a function

ihor-sviziev:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This is a follow-up on https://github.com/magento/magento2/issues/33595 and https://github.com/symfony/symfony/pull/42281.
Magento using laminas/laminas-code as a dependency. It is used especially for retrieving argument types during running the `php bin/magento setup:di:compile` command.

The root cause analyses showed that inside the module `symfony/console` was removed the optional `@param` that declared as argument types. The hotfix was the following - declare all params.

My PR changes behavior from checking only by position to two steps when NOT all arguments have their own argument type defined:
1. Check property by position (the fastest solution)
    - if the variable name is empty (e.g.`* @param string`) - use return type from it
    - if the variable name is equal to what was requested (e.g.`* @param string $myParam`) - use return type from it
2. Iterate all `@param` tags from the phpdoc block and try to find the corresponding variable name. If found - user return type from it

### DIFF
--- a/src/Reflection/ParameterReflection.php
+++ b/src/Reflection/ParameterReflection.php
@@ -2,6 +2,7 @@
 
 namespace Laminas\Code\Reflection;
 
+use Laminas\Code\Reflection\DocBlock\Tag\ParamTag;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionParameter;
@@ -92,10 +93,19 @@ class ParameterReflection extends ReflectionParameter implements ReflectionInter
             return null;
         }
 
-        $params = $docBlock->getTags('param');
+        /** @var ParamTag[] $params */
+        $params       = $docBlock->getTags('param');
+        $paramTag     = $params[$this->getPosition()] ?? null;
+        $variableName = '$' . $this->getName();
 
-        if (isset($params[$this->getPosition()])) {
-            return $params[$this->getPosition()]->getType();
+        if ($paramTag && ('' === $paramTag->getVariableName() || $variableName === $paramTag->getVariableName())) {
+            return $paramTag->getTypes()[0] ?? '';
+        }
+
+        foreach ($params as $param) {
+            if ($param->getVariableName() === $variableName) {
+                return $param->getTypes()[0] ?? '';
+            }
         }
 
         return null;
--- a/test/Reflection/ParameterReflectionTest.php
+++ b/test/Reflection/ParameterReflectionTest.php
@@ -59,11 +59,11 @@ class ParameterReflectionTest extends TestCase
     }
 
     /**
-     * This test covers
+     * This test covers type detection when not all params declared in phpDoc block
      *
-     * @dataProvider paramType2
+     * @dataProvider paramTypeWithNotAllParamsDeclared
      */
-    public function testTypeReturn2(string $param, string $type): void
+    public function testTypeReturnWithNotAllParamsDeclared(string $param, string $type): void
     {
         $parameter = new Reflection\ParameterReflection(
             [TestAsset\TestSampleClass5::class, 'methodWithNotAllParamsDeclared'],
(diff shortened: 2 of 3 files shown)


ihor-sviziev:
@Ocramius I just pushed the changes.
I'm first time contributing to laminas, so not sure if I did everything correctly.
Note: For Magento 2 we need to have 3.5.x release. Should I create a separate PR for that? Or you will port it?

IbrahimS2:
@ihor-sviziev The datatype should be checked first and if not declared then and only then PHPDocs can be checked.

ihor-sviziev:
@IbrahimS2, as you can see, it works exactly in this way. The new case is added when NOT all arguments has their own argument type defined.
I update the PR description, in order to make it more clear

ihor-sviziev:
@Ocramius, could you please review this pull request?

src/Reflection/ParameterReflection.php
```diff
@@ -92,10 +93,19 @@ public function detectType()
             return null;
         }
 
-        $params = $docBlock->getTags('param');
+        /** @var ParamTag[] $params */
+        $params       = $docBlock->getTags('param');
+        $paramTag     = $params[$this->getPosition()] ?? null;
```
Ocramius, line 98: I don't fully understand this particular line: what is going on here? 👀

ihor-sviziev: Here we're trying to get the parameter from phpdoc block by position (the original implementation worked like that).
This is equivalent of `$paramTag = isset($params[$this->getPosition()]) ? $params[$this->getPosition()] : null;`

In case if we weren't able to find it - we're trying to get the type by the variable name.

test/Reflection/ParameterReflectionTest.php
```diff
@@ -58,6 +58,20 @@ public function testTypeReturn($param, $type)
         self::assertEquals($type, $parameter->detectType());
     }
 
+    /**
+     * This test covers
```
Ocramius: What does this test cover?

test/Reflection/ParameterReflectionTest.php
```diff
@@ -58,6 +58,20 @@ public function testTypeReturn($param, $type)
...
+     *
+     * @dataProvider paramType2
+     */
+    public function testTypeReturn2(string $param, string $type): void
```
Ocramius: Is there a more meaningful way to name this test?

Ocramius (changes requested):
Besides minor adjustments, I think this is good to go

test/Reflection/ParameterReflectionTest.php
ihor-sviziev:
```suggestion
    public function testTypeReturnWithNotAllParamsDeclared(string $param, string $type): void
```

test/Reflection/ParameterReflectionTest.php
```diff
@@ -82,6 +96,21 @@ public function paramType(): array
         ];
     }
 
+    /**
+     * @return string[][]
+     * @psalm-return non-empty-list<array{non-empty-string, non-empty-string}>
+     */
+    public function paramType2(): array
```
ihor-sviziev:
```suggestion
    public function paramTypeWithNotAllParamsDeclared(): array
```

test/Reflection/ParameterReflectionTest.php
ihor-sviziev:
```suggestion
     * This test covers type detection when not all params declared in phpDoc block
```

test/Reflection/ParameterReflectionTest.php
```diff
@@ -58,6 +58,20 @@ public function testTypeReturn($param, $type)
         self::assertEquals($type, $parameter->detectType());
     }
 
+    /**
+     * This test covers
+     *
+     * @dataProvider paramType2
```
ihor-sviziev:
```suggestion
     * @dataProvider paramTypeWithNotAllParamsDeclared
```

Ocramius (approved):
:ship: thanks @ihor-sviziev!

PR #108, opened 2021-09-26
[Php74] Apply PHP 7.4 syntax and typed properties (for private modifier)

samsonasik:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

[Php74] Apply PHP 7.4 syntax and typed properties (for private modifier) to keep BC.

### DIFF
--- a/src/Generator/ClassGenerator.php
+++ b/src/Generator/ClassGenerator.php
@@ -446,10 +446,9 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
      */
     public function setImplementedInterfaces(array $implementedInterfaces)
     {
-        array_map(function ($implementedInterface) {
-            // This loop is just validating that the given `$implementedInterfaces` contains valid syntax/symbols
-            return TypeGenerator::fromTypeString($implementedInterface);
-        }, $implementedInterfaces);
+        // This loop is just validating that the given `$implementedInterfaces` contains valid syntax/symbols
+        array_map(static fn($implementedInterface) =>
+            TypeGenerator::fromTypeString($implementedInterface), $implementedInterfaces);
 
         $this->implementedInterfaces = $implementedInterfaces;
         return $this;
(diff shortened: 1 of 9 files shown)


src/Generator/ClassGenerator.php
```diff
@@ -446,10 +446,9 @@ public function removeExtentedClass()
      */
     public function setImplementedInterfaces(array $implementedInterfaces)
     {
-        array_map(function ($implementedInterface) {
-            // This loop is just validating that the given `$implementedInterfaces` contains valid syntax/symbols
-            return TypeGenerator::fromTypeString($implementedInterface);
-        }, $implementedInterfaces);
+        // This loop is just validating that the given `$implementedInterfaces` contains valid syntax/symbols
```
Ocramius: I wonder if we can just drop this?

Ocramius (approved):

src/Generator/ClassGenerator.php
```diff
@@ -446,10 +446,9 @@ public function removeExtentedClass()
...
+        array_map(static fn($implementedInterface) =>
+            TypeGenerator::fromTypeString($implementedInterface), $implementedInterfaces);
```
samsonasik: any way to suppresss this psalm error: `The call to array_map is not used` ? I tried with : 

```bash
@psalm-suppress UnusedFunctionCall
```

but it seems still error, even added to `psalm.xml`:

```xml
        <UnusedFunctionCall>
            <errorLevel type="error">
                <referencedFunction name="UnusedFunctionCall"/>
            </errorLevel>
        </UnusedFunctionCall>
```

src/Generator/ClassGenerator.php
samsonasik: I dropped it.

samsonasik:
All green 🎉

Ocramius (approved):

Ocramius:
Thanks @samsonasik!

PR #109, opened 2021-10-18
Add initial use cases for Enum generation

### LINKED ISSUE #100: PHP 8.1 enum type support
https://wiki.php.net/rfc/enumerations

agustingomes:
|    Q          |   A
|-------------- | ------
| Documentation | yes (at later stage)
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

This PR Enables the creation of Enum objects for the following use cases:

- Pure Enums
- Backed Enums either with string or integer values (as per RFC specification)

Fixes #100


src/Generator/EnumGenerator.php
```diff
@@ -0,0 +1,136 @@
...
+    private string $name = '';
+
+    private ?string $scalarType = null;
+
+    /**
+     * @var list<string>
+     */
+    private array $cases = [];
+
+    public function setName(string $name): self
```
Ocramius: On the new API, I wonder if we can leave out all setters as much as possible, and only use the constructor?

Ocramius (changes requested):

src/Generator/EnumGenerator.php
agustingomes: I was thinking the same, using something like `EnumGenerator::fromArray()` to generate a new instance that is immutable sounds more solid.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * @param array<string, mixed> $backedCases
+     */
+    public static function fromCasesWithType(array $backedCases, string $type): self
+    {
+        $cases = [];
+        foreach ($backedCases as $case => $value) {
+            if ($type === 'string') {
```
weierophinney, line 58: Add a `&& is_string($value) || (is_object($value) && method_exists($value, '__toString'))` clause, and this error and the next should go away.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,52 @@
...
+                $value = '\'' . $value . '\'';
+            }
+
+            $cases[] = $case . ' = ' . $value;
```
weierophinney, line 62: This makes it appear that the proper type for `$backedCases` is `array<string, string>`. Making that change may also correct the errors flagged i this method.

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,76 @@
...
+        return $output;
+    }
+
+    private function retrieveType(): string
+    {
+        if ($this->cases->getBackedType() === null) {
+            return '';
+        }
+
+        return ': ' . $this->cases->getBackedType();
```
weierophinney, line 64:
```suggestion
        return ': ' . $this->cases->getBackedType() ?: '';
```

src/Generator/EnumGenerator/Name.php
```diff
@@ -0,0 +1,41 @@
...
+        return $this->namespace;
+    }
+
+    public static function fromName(string $name): self
+    {
+        $namespace = null;
+        if (false !== strpos($name, '\\')) {
+            $namespace = substr($name, 0, strrpos($name, '\\'));
+            $name      = substr($name, strrpos($name, '\\') + 1);
+        }
```
weierophinney:
```suggestion
        $nsPosition = strrpos($name, '\\');
        if (false !== $nsPosition) {
            $namespace = substr($name, 0, $nsPosition);
            $name      = substr($name, $nsPosition + 1);
        }
```

This will correct the error, and prevent duplicate operations, making it perform faster.

weierophinney (changes requested):
The bulk of the static analysis errors are due to the engine needing to infer types, and not being able to do so. The way to correct that is to verify them before doing an operation. For values in arrays, check that the array key exists, and then that the value conforms to what is required, before performing an operation. In other cases, just make sure that the typehints provided are correct, particularly in annotations. If needed, uses webmozart/assert to do the validations.

agustingomes:
Thanks @weierophinney for the initial review. I'll keep your feedback in account as I work further in the PR

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: I ended solving this error by tightening the docblock to the other allowed value, in this case, `int`

phpcs.xml
```diff
@@ -9,6 +9,12 @@
     <exclude-pattern>*/TestAsset/*</exclude-pattern>
     <exclude-pattern>*/_files/*</exclude-pattern>
 
+    <rule ref="Generic.Files.LineLength.TooLong">
```
agustingomes: @Ocramius @weierophinney I added this ignore rule because the annotations for psalm I added were quite long.

weierophinney: You can split annotations over multiple lines, especially those detailing array or object structures; I tend to prefer that approach as it has the side effect of making them more readable.

agustingomes: That worked indeed. I pushed the change.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @param array<string, string|int> $backedCases
+     */
+    public static function fromCasesWithType(array $backedCases, string $type): self
+    {
+        $cases = [];
+        foreach ($backedCases as $case => $value) {
+            if ($type === 'string') {
+                $value = '\'' . $value . '\'';
```
weierophinney: I'd go with either:

```php
$value = "'" . $value . "'";
```

or:

```php
$value = sprintf("'%s'", $value);
```

to make it more clear that what you're doing is surrounding a string value with single quotes.

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if ($isPureCasesSpecified && $isBackedCasesSpecified === false) {
+            return self::fromPureCases($options['pureCases'] ?? []);
+        }
+
+        if ($isPureCasesSpecified === false && $isBackedCasesSpecified) {
+            return self::fromBackedCases($options['backedCases'] ?? ['type' => '', 'cases' => []]);
+        }
+
+        //TODO add more specific exception
```
weierophinney:
```suggestion
        /** @todo add more specific exception */
```

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
...
+            ],
+            <<<CODE
+enum Suit {
+    case Hearts;
+    case Diamonds;
+    case Clubs;
+    case Spades;
+}
+
+CODE,
```
weierophinney:
```suggestion
            <<<CODE
                enum Suit {
                    case Hearts;
                    case Diamonds;
                    case Clubs;
                    case Spades;
                }

                CODE,
```

Since PHP 7.3, heredocs and nowdocs support indentation; any whitespace preceding the end delimiter will be stripped from the start of lines in the document. It's a nice feature to make inline heredoc/nowdocs more readable

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
...
+namespace My\Namespaced;
+
+enum Suit {
+    case Hearts;
+    case Diamonds;
+    case Clubs;
+    case Spades;
+}
+
+CODE,
```
weierophinney:
```suggestion
            <<<CODE
                namespace My\Namespaced;

                enum Suit {
                    case Hearts;
                    case Diamonds;
                    case Clubs;
                    case Spades;
                }

                CODE,
```

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
...
+namespace My\Namespaced;
+
+enum Suit: string {
+    case Hearts = 'H';
+    case Diamonds = 'D';
+    case Clubs = 'C';
+    case Spades = 'S';
+}
+
+CODE,
```
weierophinney:
```suggestion
            <<<CODE
                namespace My\Namespaced;

                enum Suit: string {
                    case Hearts = 'H';
                    case Diamonds = 'D';
                    case Clubs = 'C';
                    case Spades = 'S';
                }

                CODE,
```

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
...
+namespace My\Namespaced;
+
+enum Suit: int {
+    case Hearts = 1;
+    case Diamonds = 2;
+    case Clubs = 3;
+    case Spades = 4;
+}
+
+CODE,
```
weierophinney:
```suggestion
            <<<CODE
                namespace My\Namespaced;

                enum Suit: int {
                    case Hearts = 1;
                    case Diamonds = 2;
                    case Clubs = 3;
                    case Spades = 4;
                }

                CODE,
```

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
...
+    case Diamonds;
+    case Clubs;
+    case Spades;
+
+    public function test() : string
+    {
+    }
+}
+
+CODE,
```
weierophinney:
```suggestion
            <<<CODE
                enum Suit {
                    case Hearts;
                    case Diamonds;
                    case Clubs;
                    case Spades;

                    public function test() : string
                    {
                    }
                }

                CODE,
```

weierophinney (changes requested):
Few minor suggestions, mainly for readability. After making them, I think you can go ahead and remove the draft status, and @Ocramius can do final review.

test/Generator/EnumGeneratorTest.php
agustingomes: Aah that's cool, it's definitely much better the way you're suggesting.

agustingomes:
@weierophinney Thank you for the suggestions. I marked the PR as ready for review.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,54 @@
...
+
+final class BackedCases implements CasesInterface
+{
+    private string $type;
+
+    /** @var list<string> */
+    private array $cases;
+
+    /**
+     * @param list<string> $cases
```
Ocramius: Are these `string` or `non-empty-string`?

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Laminas\Code\Generator\EnumGenerator\Cases;
+
+use Laminas\Code\Generator\EnumGenerator\CasesInterface;
+
+use function sprintf;
+
+final class BackedCases implements CasesInterface
```
Ocramius: Let's mark this `/** @psalm-immutable */`

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+use Laminas\Code\Generator\EnumGenerator\CasesInterface;
+
+use function array_key_exists;
+
+final class CaseFactory
+{
+    /**
+     * @psalm-param array{
+     *      name: string,
+     *      pureCases?: list<string>,
```
Ocramius: Are these `string` or `non-empty-string`?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     *      backedCases?: array{
+     *          type: string,
+     *          cases: array<string, int|string>
```
Ocramius: Are these `string` or `non-empty-string`?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     *      },
+     *      methods?: list<array{
+     *          name: string,
+     *          visibility?: string,
```
Ocramius: `visibility` as `string` seems a bit weak - can we improve on this, and document `'public'|'protected'|'private'`?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @psalm-param array{
+     *      name: string,
+     *      pureCases?: list<string>,
+     *      backedCases?: array{
+     *          type: string,
+     *          cases: array<string, int|string>
+     *      },
+     *      methods?: list<array{
+     *          name: string,
```
Ocramius: Is this a `string` or `non-empty-string`?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     *          visibility?: string,
+     *          static?: bool,
+     *          returntype?: string,
```
Ocramius: About `returntype`:

 * `returntype` or `returnType`?
 * could we expect a `TypeGenerator` instead of `string`?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     *      }>
+     * } $options
+     */
+    public static function fromOptions(array $options): CasesInterface
+    {
+        $isPureCasesSpecified   = array_key_exists('pureCases', $options);
+        $isBackedCasesSpecified = array_key_exists('backedCases', $options);
+
+        if ($isPureCasesSpecified && $isBackedCasesSpecified === false) {
```
Ocramius: `$isBackedCasesSpecified === false` should imo be `! $isBackedCasesSpecified`, but it's just stylistic

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+            return self::fromPureCases($options['pureCases'] ?? []);
```
Ocramius: The `?? []` should not be needed here - we just verified that `array_key_exists()`

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if ($isPureCasesSpecified === false && $isBackedCasesSpecified) {
```
Ocramius: `$isPureCasesSpecified === false` should be `! $isPureCasesSpecified` IMO

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+            return self::fromBackedCases($options['backedCases'] ?? ['type' => '', 'cases' => []]);
+        }
+
+        /** @todo add more specific exception */
```
Ocramius: To be solved before merge?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+        throw new Exception('Either `pureCases` or `backedCases` must be specified.');
```
Ocramius: We can avoid this scenario at type level.

The docblock would need to be exploded a bit:

```php
    /**
     * @psalm-param array{
     *      name: string,
     *      pureCases: list<string>,
     *      methods?: list<array{
     *          name: string,
     *          visibility?: string,
     *          static?: bool,
     *          returntype?: string,
     *      }>
     * }|array{
     *      name: string,
     *      backedCases: array{
     *          type: string,
     *          cases: array<string, int|string>
     *      },
     *      methods?: list<array{
     *          name: string,
     *          visibility?: string,
     *          static?: bool,
     *          returntype?: string,
     *      }>
     * } $options
     */
```

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @param list<string> $pureCases
```
Ocramius: `non-empty-string` ?

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public static function fromPureCases(array $pureCases): PureCases
+    {
+        return PureCases::fromCases($pureCases);
+    }
+
+    /**
+     * @psalm-param array{
+     *     type: string,
```
Ocramius: We should probably use `TypeGenerator` here

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,63 @@
...
+     *     cases: array<string, int|string>
+     * } $backedCases
+     */
+    public static function fromBackedCases(array $backedCases): BackedCases
+    {
+        return BackedCases::fromCasesWithType($backedCases['cases'] ?? [], $backedCases['type'] ?? '');
```
Ocramius: `$backedCases['type'] ?? ''` contradicts the declared type

Ocramius: `$backedCases['cases'] ?? []` contradicts the declared type

src/Generator/EnumGenerator/Cases/PureCases.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Laminas\Code\Generator\EnumGenerator\Cases;
+
+use Laminas\Code\Generator\EnumGenerator\CasesInterface;
+
+use function array_values;
+
+final class PureCases implements CasesInterface
+{
+    /** @var list<string> */
```
Ocramius: `non-empty-string`?

src/Generator/EnumGenerator/CasesInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Laminas\Code\Generator\EnumGenerator;
+
+interface CasesInterface
```
Ocramius: This looks like a sealed type, where `CasesInterface = PureCases|BackedCases`

I suggest getting rid of the interface, and declaring the two types directly: this prevents implementing `CasesInterface` in userland, which would simplify our assumptions quite a bit.

src/Generator/EnumGenerator/Cases/PureCases.php
```diff
@@ -0,0 +1,42 @@
...
+     * @param list<string> $cases
+     */
+    private function __construct(array $cases)
+    {
+        $this->cases = $cases;
+    }
+
+    public function getBackedType(): ?string
+    {
+        return null;
```
Ocramius: If we remove the interface (as suggested below), and only use `PureCases|BackedCases` in our system, this nullability goes away

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,146 @@
...
+
+    public function generate(): string
+    {
+        $output = '';
+
+        if (null !== $this->name->getNamespace()) {
+            $output .= 'namespace ' . $this->name->getNamespace() . ';' . self::LINE_FEED . self::LINE_FEED;
+        }
+
+        $output .= 'enum ' . $this->name->getName() . $this->retrieveType() . ' {'
```
Ocramius: Can return directly

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,146 @@
...
+                fn (string $case): string => self::INDENTATION . 'case ' . $case . ';' . self::LINE_FEED,
+                $this->cases->getCases()
+            )
+        );
+    }
+
+    /**
+     * @psalm-param array{
+     *      name: string,
+     *      pureCases?: list<string>,
```
Ocramius: The declaration of one of either `pureCases` or `backedCases` needing to exist should be ported here too, as shown in the `CaseFactory` suggestions

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,146 @@
...
+     *      methods?: list<array{
+     *          name: string,
+     *          visibility?: string,
+     *          static?: bool,
+     *          returntype?: string,
+     *      }>
+     * } $options
+     * @return list<MethodGenerator>
+     */
+    private static function fromMethods(array $options): array
```
Ocramius: This is not creating a `:self`, so the naming `::fromMethods()` is a bit confusing :thinking:

src/Generator/EnumGenerator/Name.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Laminas\Code\Generator\EnumGenerator;
+
+use function strpos;
+use function strrpos;
+use function substr;
+
+final class Name
```
Ocramius, line 13: Should probably be `/** @internal */` and have `/** @psalm-immutable */`

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,146 @@
...
+namespace Laminas\Code\Generator\EnumGenerator;
+
+use Laminas\Code\Generator\EnumGenerator\Cases\CaseFactory;
+use Laminas\Code\Generator\MethodGenerator;
+
+use function array_key_exists;
+use function array_map;
+use function implode;
+
+final class EnumGenerator
```
Ocramius, line 17: This could potentially have a `/** @psalm-immutable */` attached

src/Generator/EnumGenerator/EnumGenerator.php
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        $methodsOutput = '';
+        foreach ($this->methods as $method) {
+            $methodsOutput .= self::LINE_FEED . $method->generate();
+        }
+
+        return $methodsOutput;
+    }
+}
```
Ocramius, line 109: One method missing here is `::fromReflection()`, operating on a given enum reflection instance - that's one of the biggest utilities in this generator system.

src/Generator/EnumGenerator/Cases/PureCases.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Laminas\Code\Generator\EnumGenerator\Cases;
+
+use Laminas\Code\Generator\EnumGenerator\CasesInterface;
+
+use function array_values;
+
+final class PureCases implements CasesInterface
```
Ocramius: `/** @psalm-immutable */`

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+namespace LaminasTest\Code\Generator;
+
+use Laminas\Code\Generator\EnumGenerator\EnumGenerator;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @group Laminas_Code_Generator
```
Ocramius: These groups can probably be dropped

src/Generator/EnumGenerator/Cases/BackedCases.php
Ocramius: If I understand it correctly, `BackedCases` and `PureCases` are internal details, and should be marked `/** @internal */`?

Ocramius (changes requested):

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: `non-empty-strings` is better here.

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: That is correct.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: I tend to prefer adding the explicit comparison, as the `!` could be easily missed by the code readers at first glance.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: Psalm complains if I omit that addition, and I'm not sure if it's because it's not understanding the check we're performing above.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: Yes, I'll add a new exception. Do you have any suggestions what could we name it? `InvalidCasesSpecificationException` maybe?

src/Generator/EnumGenerator/Cases/CaseFactory.php
Ocramius: Potentially to be written as follows then (without variable):

```php
if (array_key_exists('backedCases', $options) ...) {
```

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: But API consumers could still specify both properties in their implementations when using `EnumGenerator::fromConfig()`, wouldn't it be safer to throw the exception if that happens?

Even after implementing this dockblock, the analysis complains that the offsets of `pureCases` and `backedCases` could be undefined in their respective lines, even though we are checking beforehand.

src/Generator/EnumGenerator/CasesInterface.php
agustingomes: I agree, but that would require bumping the minimum required PHP version to 8.0, Should we do this in this PR?

src/Generator/EnumGenerator/EnumGenerator.php
agustingomes: would `::withProvidedMethodOptions()` work better?

src/Generator/EnumGenerator/Cases/CaseFactory.php
Ocramius: See comments around this: if we refine the types, the exception is not needed

src/Generator/EnumGenerator/Cases/CaseFactory.php
Ocramius: > could be undefined in their respective lines, even though we are checking beforehand.

Yes, that is because type inference can't understand if the left or right trees (in the suggested type above) are used.

That could be an `assert(array_key_exists(...))` after the first conditional:

```php
if (array_key_exists(...)) {
    return ...;
}

assert(array_key_exists(...));
return ...;
```

src/Generator/EnumGenerator/EnumGenerator.php
Ocramius: IMO no `with` in the naming: I would expect a `: self` return type, if it was called `with*` :) 

`private static function makeMethods(...): array` or such would fit better

src/Generator/EnumGenerator/EnumGenerator.php
agustingomes: Unfortunately we can't mark it as immutable because it's depending on the `MethodGenerator` and that class is not immutable :'(

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: that did the trick

src/Generator/EnumGenerator/CasesInterface.php
agustingomes: nevermind, doing it with DocBlocks is also good.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: the property in the `MethodGenerator` is not camelcase, hence I left it like this.

I'll check the TypeGenerator suggestion.

Edit: the method setting the return type does not accept the `TypeGenerator` method. Is this something we have to change now, or can it be done in a separate PR?
image

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: I tried to remove the exception thrown, but no luck. since it's possible for people in userland wrongly send options without neither property set, it seems we need to throw the exception when it happens. Do you see another way around this?

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: indeed, moving them inline worked.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: never mind, using the `assert` above the return of the `backedCases` worked.

src/Generator/EnumGenerator/EnumGenerator.php
agustingomes: I'll work on adding this method, but how can we do that while remaining compatible with 7.4?

From what I understand in the RFC, the reflection we need to use for that is `ReflectionEnum` but that is only included in 8.1.

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: would `non-empty-string` work instead?

Ocramius: > Edit: the method setting the return type does not accept the TypeGenerator method. Is this something we have to change now, or can it be done in a separate PR?

IMO we shouldn't expand more `type: string` API: it is a long-term liability.

Alternatively, completely drop this API, and remove the ability to use `::fromArray()` for this stuff.

agustingomes: Yeah, I feel it's a good compromise. I'll drop the methods generation from this PR.

src/Generator/EnumGenerator/Name.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class Name
+{
+    private string $name;
+
+    private ?string $namespace;
```
michaelpetri, line 17: Would both expect to be non-empty-string

src/Generator/EnumGenerator/Name.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function getNamespace(): ?string
+    {
+        return $this->namespace;
+    }
```
michaelpetri, line 33: Since class is immutable, we can drop this.

src/Generator/EnumGenerator/Name.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public static function fromName(string $name): self
```
michaelpetri: Would rename to `fromFqcn`

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public function getBackedType(): ?string
+    {
+        return $this->type;
+    }
+
+    /**
+     * @inheritDoc
+     */
```
michaelpetri: Class doesn't inherit from anything. But i wonder if it would make sense to introduce one in order to have a real type [here](https://github.com/laminas/laminas-code/pull/109/files#diff-a5540f944bc28e9bb26f0d15a7475d2f83b42347b9301f4b17bc9788ba2e104aR28) and [here](https://github.com/laminas/laminas-code/pull/109/files#diff-a5540f944bc28e9bb26f0d15a7475d2f83b42347b9301f4b17bc9788ba2e104aR33)

src/Generator/EnumGenerator/EnumGenerator.php
michaelpetri: Just an idea: https://3v4l.org/2SYBR

src/Generator/EnumGenerator/Cases/CaseFactory.php
```diff
@@ -0,0 +1,33 @@
...
+     * }|array{
+     *      name: non-empty-string,
+     *      backedCases: array{
+     *          type: 'int'|'string',
+     *          cases: array<non-empty-string, int|non-empty-string>,
+     *      },
+     * } $options
+     * @return BackedCases|PureCases
+     */
+    public static function fromOptions(array $options)
```
michaelpetri, line 35: Related to the interface comment. But maybe instead of an interface we could introduce a abstract class and merge it with the factory:

```php
<?php

/** @internal */
abstract class Cases {
    /** @var list<non-empty-string> */
    private array $cases;

    /** @param list<non-empty-string> $cases */
    protected function __construct(array $cases)
    {
        $this->cases = $cases;
    }

    /** @psalm-return list<non-empty-string> */
    public function getCases(): array
    {
        return $this->cases;
    }

    /**
     * @psalm-param array{
     *      name: non-empty-string,
     *      pureCases: list<non-empty-string>,
     * }|array{
     *      name: non-empty-string,
     *      backedCases: array{
     *          type: 'int'|'string',
     *          cases: array<non-empty-string, int|non-empty-string>,
     *      },
     * } $options
     * @return BackedCases|PureCases
     */
    final public static function fromOptions(array $options): self
    {
        if (array_key_exists('pureCases', $options) && ! array_key_exists('backedCases', $options)) {
            return PureCases::fromCases($options['pureCases']);
        }

        assert(! array_key_exists('pureCases', $options) && array_key_exists('backedCases', $options));
        return BackedCases::fromCasesWithType($options['backedCases']['cases'], $options['backedCases']['type']);
    }
}

/**
 * @internal
 * @psalm-immutable
 */
final class PureCases extends Cases
{
    /** @psalm-param list<non-empty-string> $pureCases */
    public static function fromCases(array $pureCases): self
    {
        return new self($pureCases);
    }
}

/** 
 * @internal
 * @psalm-immutable  
 */
final class BackedCases extends cases
{
    private string $type;

    /** @param list<non-empty-string> $cases */
    private function __construct(string $type, array $cases)
    {
        $this->type  = $type;
        parent::__construct($cases);
    }

    public function getBackedType(): ?string
    {
        return $this->type;
    }

    /** @param array<string, string|int> $backedCases */
    public static function fromCasesWithType(array $backedCases, string $type): self
    {
        $cases = [];
        foreach ($backedCases as $case => $value) {
            if ($type === 'string') {
                $value = sprintf("'%s'", $value);
            }

            $cases[] = $case . ' = ' . $value;
        }

        return new self($type, $cases);
    }
}
```

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,57 @@
...
+    public function getCases(): array
+    {
+        return $this->cases;
+    }
+
+    /**
+     * @param array<string, string|int> $backedCases
```
michaelpetri:
```suggestion
     * @param array<string, int>|array<string, string> $backedCases
```
Cases can't be mixed, Same [here](https://github.com/laminas/laminas-code/pull/109/files#diff-e301a27a5ae98f04f13cad35f3b4f5d41838e886deb3e0db9779fdc60967e84aR19) 

Also I really wonder if we need the second parameter when we know this?

src/Generator/EnumGenerator/Name.php
agustingomes: I'm not sure what do you mean here, you mean to remove the getters and make the properties public?

src/Generator/EnumGenerator/Name.php
agustingomes: Existing implementations allow namespace to be omitted, so I assume we want to retain that characteristic for Enums.

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: Thanks for pointing it, will remove on the next push.

src/Generator/EnumGenerator/EnumGenerator.php
agustingomes: I'll take a look at the suggestion

src/Generator/EnumGenerator/Cases/CaseFactory.php
agustingomes: Personally I try to avoid abstract classes whenever possible, and for this particular case, I think putting the "factory" as part of the abstract class is not very extensible.

Next year, when PHP 7.4 is dropped (assuming that's the plan) we can use Union types instead.

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: I'll add the suggestion, thanks 👍🏼 

As for the second param, I think having it is better than inferring the type from the array values.

src/Generator/EnumGenerator/Name.php
agustingomes: if it's deemed we must have a fully qualified class name, I'll adopt this suggestion.

src/Generator/EnumGenerator/EnumGenerator.php
agustingomes: I implemented the `::fromReflection()` method.

I updated the Psalm baseline for now, since I have no knowledge how to tackle the specific issues that arose from it, especially the `ParseError` ones, which makes sense since the analysis is done in 8.0.

src/Generator/EnumGenerator/Name.php
michaelpetri: Yes, sorry what i mean is `null|non-empty-string`.

src/Generator/EnumGenerator/Name.php
michaelpetri: Yes exactly

src/Generator/EnumGenerator/Name.php
michaelpetri: I just think its a bit confusing since a property `name` exists, one could assume that the named ctor only sets the name. And since the fqcn of a class without a namespace is the class name itself it makes at least sense to me. But if you disagree just ignore it.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @param list<non-empty-string> $cases
+     */
+    private function __construct(string $type, array $cases)
+    {
+        $this->type  = $type;
+        $this->cases = $cases;
+    }
+
+    public function getBackedType(): ?string
```
weierophinney: This will never be `null`, per the constructor and the declared property type.

src/Generator/EnumGenerator/Cases/BackedCases.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public function getCases(): array
+    {
+        return $this->cases;
+    }
+
+    /**
+     * @param array<string, int>|array<string, string> $backedCases
+     */
+    public static function fromCasesWithType(array $backedCases, string $type): self
```
weierophinney, line 47: We need validation on `$type`. Enums only allow for int or string types when using backed types.

src/Generator/EnumGenerator/Name.php
weierophinney: I'd wait until the minimum supported version on this is 8.1, and we can mark the properties readonly. Until then, it makes sense to have the properties private and use getters, as it prevents users from overwriting the values.

src/Generator/EnumGenerator/Name.php
weierophinney: I agree with `fromFqcn()` or `fromFullyQualifiedClassName()`. A global Enum with no namespace is fully qualified already, and hence would have no nuamespace associated; namespaced Enums will be split into a namespace + name. The name of the method should reflect this.

weierophinney (changes requested):
Getting close now! Thanks for all the work you're doing on this!

src/Generator/EnumGenerator/Cases/BackedCases.php
agustingomes: Will also add the respective exception here.

src/Generator/EnumGenerator/Name.php
agustingomes: I agree with @weierophinney 

In theory, any downstream project could abuse this object regardless that is annotated as `internal`. waiting for PHP 8.1 to become the minimum sounds good for me.

src/Generator/EnumGenerator/Name.php
agustingomes: I'll adopt the longer option since it's more explicit.

src/Generator/EnumGenerator/Name.php
agustingomes: Since that property can still be `null`, I'll skip this suggestion.

weierophinney (approved):
:+1:

Will wait for @Ocramius to do final review, however.

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,207 @@
...
+                
+                CODE,
+        ];
+    }
+
+    public function testReflectionEnumFailsForUnsupportedPhpVersions(): void
+    {
+        if (PHP_VERSION_ID >= 80100) {
+            $this->markTestSkipped('ReflectionEnum is available from PHP 8.1 onwards.');
+        }
```
michaelpetri: You could use `/** @requires PHP >= 8.1 */` on test methods

test/Generator/EnumGeneratorTest.php
```diff
@@ -0,0 +1,207 @@
...
+        EnumGenerator::fromReflection(new ReflectionEnum(new class {
+        }));
+    }
+
+    /**
+     * @dataProvider validEnumSpecifications
+     * @psalm-param array{
+     *      0: class-string,
+     *      1: non-empty-string
+     * }
```
michaelpetri:
```suggestion
     * @psalm-param class-string $enumClass
     * @psalm-param non-empty-string expected
```

psalm-baseline.xml
```diff
@@ -1,5 +1,22 @@
...
+      <code>Flags</code>
+      <code>case</code>
+    </ParseError>
+  </file>
+  <file src="polyfill/test/Orientation.php">
+    <ParseError occurrences="2">
+      <code>Orientation</code>
+      <code>case</code>
+    </ParseError>
+  </file>
```
michaelpetri: For other 8.1 features i used eval to generate classes while runtime, see https://github.com/laminas/laminas-code/blob/4.5.x/test/Generator/PropertyGeneratorTest.php#L405

psalm-baseline.xml
```diff
@@ -258,6 +274,47 @@
...
+    </UndefinedMethod>
+  </file>
+  <file src="src/Generator/EnumGenerator/EnumGenerator.php">
+    <MixedArgument occurrences="1">
+      <code>$enum-&gt;getName()</code>
+    </MixedArgument>
+    <UndefinedMethod occurrences="1">
+      <code>getName</code>
+    </UndefinedMethod>
+  </file>
```
michaelpetri, line 300: Not sure if worth to do but you could extend the polyfill to get rid of this.

Ocramius (approved):
Overall excellent:

 * the fact that implementations are `@internal` gives us space to improve later
 * everything else is simple and to the point

I'll proceed to merge, and will factor in the last comments by @michaelpetri  ( https://github.com/laminas/laminas-code/pull/109#pullrequestreview-824445421 )

psalm-baseline.xml
Ocramius: Yeah, these will need to be moved to strings, for now: reason being that psalm chokes in a clean environment too.

psalm-baseline.xml
Ocramius: IMO not worth doing here - these will disappear as we increase minimum PHP version, when viable.

psalm-baseline.xml
Ocramius: 5e6e3d013c32b3681533143b8dc7bc27fcd25d1d

test/Generator/EnumGeneratorTest.php
Ocramius: 5e6e3d013c32b3681533143b8dc7bc27fcd25d1d

test/Generator/EnumGeneratorTest.php
Ocramius: 5e6e3d013c32b3681533143b8dc7bc27fcd25d1d

Ocramius:
:ship: 

Thanks @agustingomes @michaelpetri @weierophinney! Gonna cut a release after dependency upgrades

PR #110, opened 2021-10-20
Support php 8

4c0n:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

I need PHP 8 support in this version as it is being used as a sub dependency in one of my projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "extra": {
     },
     "require": {
-        "php": "^7.1",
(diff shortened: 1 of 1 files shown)


Ocramius:
We already support PHP 8:

https://github.com/laminas/laminas-code/blob/4.0.0/composer.json#L23

PR #111, opened 2021-10-20
Add php 8 support to 3.4

4c0n:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR adds PHP 8 support to the 3.4 version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "extra": {
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.1 || ^8.0",
         "laminas/laminas-eventmanager": "^2.6 || ^3.0",
         "laminas/laminas-zendframework-bridge": "^1.0"
     },


Ocramius:
Duplicate of #110 - `3.4.x` is for bugfix-only: please upgrade.

4c0n:
@Ocramius So this is a sub dependency, so I cannot just upgrade to a new major version. Guess I'll have to lobby at the other project then, however there is little to no chance that they will respond... So then I'm stuck with a fork for this tiny change...

4c0n:
https://github.com/Sylius/Sylius/blob/49b070933b32761de08475fafd6d86dca10d40ff/composer.json#L164

Ocramius:
@4c0n if sylius has issues with `laminas/laminas-code:^4`, that really needs resolution here: the problem shouldn't be propagated to other dependencies.

The source of the problem seems to be https://github.com/Sylius/Sylius/commit/f80f42d06ffb98a8a5ffaaf543ccd646cdd202c6, and I suggest talking to @pamil directly: they are quite responsive :-) 

It's most likely fixable there, but asking other projects to increase dependency ranges is not viable, especially if not accompanied with extensive testing of those dependencies, which I've done @ https://github.com/laminas/laminas-code/compare/3.5.1...4.0.0 (should give you an idea of the extent of work put in it).

The code in `3.x` does **not** run against PHP 8 reflected symbols: it will just crash in bizarre ways, so it will "install", but ((it won't really be compatible** with 8.0+.

4c0n:
@Ocramius Thanks for your elaboration, it helps clear a lot up. Unfortunately Kamil is no longer a part of the Sylius team and lately they have not been very responsive at all. I don't know why exactly, but using my fork on PHP8 is not crashing at all, maybe it's just a dependency that is going unused then.
I guess I'll keep it for now and hopefully someone at the Sylius team will have a look at it and respond.

Thanks again!

Ocramius:
Keeping a fork while the issue is being fixed upstream is certainly a workaround.

You can also try a `"require": {"laminas/laminas-code": "4.4.3 as 3.99.99"}` to trick `composer` into thinking it's compatible.

PR #112, opened 2021-11-12
Added support for readonly properties

### LINKED ISSUE #101: Support PHP 8.1 read-only properties
https://wiki.php.net/rfc/readonly_properties_v2

michaelpetri:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Requested with https://github.com/laminas/laminas-code/issues/101
Fixes #101

### DIFF
--- a/src/Generator/AbstractMemberGenerator.php
+++ b/src/Generator/AbstractMemberGenerator.php
@@ -157,7 +157,7 @@ abstract class AbstractMemberGenerator extends AbstractGenerator
     }
 
     /**
-     * @return string
+     * @psalm-return static::VISIBILITY_*
      */
     public function getVisibility()
     {
--- a/src/Generator/PropertyGenerator.php
+++ b/src/Generator/PropertyGenerator.php
@@ -4,6 +4,12 @@ namespace Laminas\Code\Generator;
 
 use Laminas\Code\Reflection\PropertyReflection;
 
+use function array_reduce;
+use function get_class;
+use function gettype;
+use function is_bool;
+use function is_object;
+use function method_exists;
 use function sprintf;
 use function str_replace;
 use function strtolower;
@@ -11,6 +17,7 @@ use function strtolower;
 class PropertyGenerator extends AbstractMemberGenerator
 {
     public const FLAG_CONSTANT = 0x08;
+    public const FLAG_READONLY = 0x80;
 
     protected bool $isConst = false;
 
@@ -18,9 +25,7 @@ class PropertyGenerator extends AbstractMemberGenerator
 
     private bool $omitDefaultValue = false;
 
-    /**
-     * @return static
-     */
+    /** @return static */
     public static function fromReflection(PropertyReflection $reflectionProperty)
     {
         $property = new static();
@@ -41,6 +46,10 @@ class PropertyGenerator extends AbstractMemberGenerator
 
         if ($reflectionProperty->isStatic()) {
             $property->setStatic(true);
+        }
+
+        if (method_exists($reflectionProperty, 'isReadonly') && $reflectionProperty->isReadonly()) {
+            $property->setReadonly(true);
         }
 
         if ($reflectionProperty->isPrivate()) {
@@ -68,6 +77,7 @@ class PropertyGenerator extends AbstractMemberGenerator
      * @configkey static             bool
      * @configkey visibility         string
      * @configkey omitdefaultvalue   bool
+     * @configkey readonly           bool
      * @throws Exception\InvalidArgumentException
      * @param  array $array
      * @return static
@@ -112,6 +122,20 @@ class PropertyGenerator extends AbstractMemberGenerator
                 case 'omitdefaultvalue':
                     $property->omitDefaultValue($value);
                     break;
+                case 'readonly':
+                    if (! is_bool($value)) {
+                        throw new Exception\InvalidArgumentException(sprintf(
+                            '%s is expecting boolean on key %s. Got %s',
+                            __METHOD__,
+                            $name,
+                            is_object($value)
+                                ? get_class($value)
+                                : gettype($value)
+                        ));
+                    }
+
+                    $property->setReadonly($value);
+                    break;
             }
         }
 
@@ -119,12 +143,13 @@ class PropertyGenerator extends AbstractMemberGenerator
     }
 
     /**
-     * @param ?string $name
      * @param PropertyValueGenerator|string|array|null $defaultValue
-     * @param int $flags
+     * @param int|int[] $flags
      */
-    public function __construct($name = null, $defaultValue = null, $flags = self::FLAG_PUBLIC)
+    public function __construct(?string $name = null, $defaultValue = null, $flags = self::FLAG_PUBLIC)
     {
+        parent::__construct();
+
         if (null !== $name) {
             $this->setName($name);
         }
@@ -142,12 +167,12 @@ class PropertyGenerator extends AbstractMemberGenerator
      */
     public function setConst($const)
     {
-        if ($const) {
+        if (true === $const) {
             $this->setFlags(self::FLAG_CONSTANT);
-        } else {
-            $this->removeFlag(self::FLAG_CONSTANT);
+            return $this;
         }
 
+        $this->removeFlag(self::FLAG_CONSTANT);
         return $this;
     }
 
@@ -159,11 +184,47 @@ class PropertyGenerator extends AbstractMemberGenerator
         return (bool) ($this->flags & self::FLAG_CONSTANT);
     }
 
+    public function setReadonly(bool $readonly): self
+    {
+        if (true === $readonly) {
+            $this->setFlags(self::FLAG_READONLY);
+            return $this;
+        }
+
+        $this->removeFlag(self::FLAG_READONLY);
+        return $this;
+    }
+
+    public function isReadonly(): bool
+    {
+        return (bool) ($this->flags & self::FLAG_READONLY);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function setFlags($flags)
+    {
+        $flags = array_reduce((array) $flags, static function (int $a, int $b): int {
+            return $a | $b;
+        }, 0);
+
+        if ($flags & self::FLAG_READONLY && $flags & self::FLAG_STATIC) {
+            throw new Exception\RuntimeException('Modifier "readonly" in combination with "static" not permitted.');
+        }
+
+        if ($flags & self::FLAG_READONLY && $flags & self::FLAG_CONSTANT) {
+            throw new Exception\RuntimeException('Modifier "readonly" in combination with "constant" not permitted.');
+        }
+
+        return parent::setFlags($flags);
+    }
+
     /**
      * @param PropertyValueGenerator|mixed $defaultValue
      * @param string                       $defaultValueType
      * @param string                       $defaultValueOutputMode
-     * @return $this
+     * @return static
      */
     public function setDefaultValue(
         $defaultValue,
@@ -190,6 +251,7 @@ class PropertyGenerator extends AbstractMemberGenerator
     /**
      * @throws Exception\RuntimeException
      * @return string
+     * @psalm-return non-empty-string
      */
     public function generate()
     {
@@ -215,7 +277,11 @@ class PropertyGenerator extends AbstractMemberGenerator
                 . ($defaultValue !== null ? $defaultValue->generate() : 'null;');
         }
 
-        $output .= $this->indentation . $this->getVisibility() . ($this->isStatic() ? ' static' : '') . ' $' . $name;
+        $output .= $this->indentation
+            . $this->getVisibility()
+            . ($this->isReadonly() ? ' readonly' : '')
+            . ($this->isStatic() ? ' static' : '')
+            . ' $' . $name;
 
         if ($this->omitDefaultValue) {
             return $output . ';';
--- a/test/Generator/PropertyGeneratorTest.php
+++ b/test/Generator/PropertyGeneratorTest.php
@@ -12,12 +12,14 @@ use Laminas\Code\Generator\ValueGenerator;
 use Laminas\Code\Reflection\ClassReflection;
 use Laminas\Code\Reflection\PropertyReflection;
 use LaminasTest\Code\Generator\TestAsset\ClassWithTypedProperty;
+use PHP_CodeSniffer\Tokenizers\PHP;
 use PHPUnit\Framework\TestCase;
 use ReflectionProperty;
 use stdClass;
 
 use function array_shift;
 use function str_replace;
+use function uniqid;
 
 /**
  * @group Laminas_Code_Generator
@@ -141,9 +143,46 @@ EOS;
     }
 
     /**
+     * @dataProvider visibility
+     */
+    public function testPropertyCanProduceReadonlyModifier(int $flag, string $visibility): void
+    {
+        $codeGenProperty = new PropertyGenerator(
+            'someVal',
+            'some string value',
+            PropertyGenerator::FLAG_READONLY | $flag
+        );
+
+        self::assertSame(
+            '    ' . $visibility . ' readonly $someVal = \'some string value\';',
+            $codeGenProperty->generate()
+        );
+    }
+
+    public function testFailToProduceReadonlyStatic(): void
+    {
+        $codeGenProperty = new PropertyGenerator('someVal', 'some string value');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Modifier "readonly" in combination with "static" not permitted.');
+
+        $codeGenProperty->setFlags(PropertyGenerator::FLAG_READONLY | PropertyGenerator::FLAG_STATIC);
+    }
+
+    public function testFailToProduceReadonlyConstant(): void
+    {
+        $codeGenProperty = new PropertyGenerator('someVal', 'some string value');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Modifier "readonly" in combination with "constant" not permitted.');
+
+        $codeGenProperty->setFlags(PropertyGenerator::FLAG_READONLY | PropertyGenerator::FLAG_CONSTANT);
+    }
+
+    /**
      * @group PR-704
      */
-    public function testPropertyCanProduceContstantModifierWithSetter(): void
+    public function testPropertyCanProduceConstantModifierWithSetter(): void
     {
         $codeGenProperty = new PropertyGenerator('someVal', 'some string value');
         $codeGenProperty->setConst(true);
@@ -250,6 +289,7 @@ EOS;
 
         self::assertSame('SampleProperty', $propertyGenerator->getName());
         self::assertFalse($propertyGenerator->isConst());
+        self::assertFalse($propertyGenerator->isReadonly());
         self::assertInstanceOf(ValueGenerator::class, $propertyGenerator->getDefaultValue());
         self::assertInstanceOf(DocBlockGenerator::class, $propertyGenerator->getDocBlock());
         self::assertTrue($propertyGenerator->isAbstract());
@@ -263,6 +303,22 @@ EOS;
         $reflectionOmitDefaultValue->setAccessible(true);
 
         self::assertTrue($reflectionOmitDefaultValue->getValue($propertyGenerator));
+    }
+
+    public function testCreateReadonlyFromArray(): void
+    {
+        $propertyGenerator = PropertyGenerator::fromArray([
+            'name'     => 'ReadonlyProperty',
+            'readonly' => true,
+        ]);
+
+        self::assertSame('ReadonlyProperty', $propertyGenerator->getName());
+        self::assertFalse($propertyGenerator->isConst());
+        self::assertFalse($propertyGenerator->isAbstract());
+        self::assertFalse($propertyGenerator->isFinal());
+        self::assertFalse($propertyGenerator->isStatic());
+        self::assertTrue($propertyGenerator->isReadonly());
+        self::assertSame(PropertyGenerator::VISIBILITY_PUBLIC, $propertyGenerator->getVisibility());
     }
 
     /**
@@ -329,5 +385,20 @@ EOS;
         $code      = $generator->generate();
 
         self::assertSame('    private $typedProperty;', $code);
+    }
+
+    /** @requires PHP >= 8.1 */
+    public function testFromReflectionReadonlyProperty(): void
+    {
+        $className = uniqid('ClassWithReadonlyProperty', false);
+
+        eval('namespace ' . __NAMESPACE__ . '; class ' . $className . '{ public readonly string $readonly; }');
+
+        $reflectionProperty = new PropertyReflection(__NAMESPACE__ . '\\' . $className, 'readonly');
+
+        $generator = PropertyGenerator::fromReflection($reflectionProperty);
+        $code      = $generator->generate();
+
+        self::assertSame('    public readonly $readonly;', $code);
     }
 }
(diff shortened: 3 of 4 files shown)


src/Generator/PropertyGenerator.php
```diff
@@ -159,6 +171,47 @@ public function isConst()
         return (bool) ($this->flags & self::FLAG_CONSTANT);
     }
 
+    /**
+     * @param  bool $readonly
+     * @return PropertyGenerator
```
Ocramius: Since we are on PHP 7.4+, we can declare these types natively (with `self` and `bool`) - yes, aware that the other methods don't have that :)

src/Generator/PropertyGenerator.php
```diff
@@ -159,6 +171,47 @@ public function isConst()
...
+     */
+    public function setReadonly($readonly)
+    {
+        if ($readonly) {
+            $this->setFlags(self::FLAG_READONLY);
+        } else {
```
Ocramius: I suggest using an early return in general

src/Generator/PropertyGenerator.php
```diff
@@ -159,6 +171,47 @@ public function isConst()
...
+     */
+    public function isReadonly()
+    {
+        return (bool) ($this->flags & self::FLAG_READONLY);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function setFlags($flags)
```
Ocramius, line 206: Flags are quite a low-level representation of this: let's make this method `private`, if possible

src/Generator/PropertyGenerator.php
```diff
@@ -159,6 +171,47 @@ public function isConst()
...
+
+        if ($flags & self::FLAG_READONLY && $flags & self::FLAG_STATIC) {
+            throw new RuntimeException('Modifier "readonly" in combination with "static" not permitted.');
+        }
+
+        if ($flags & self::FLAG_READONLY && $flags & self::FLAG_CONSTANT) {
+            throw new RuntimeException('Modifier "readonly" in combination with "constant" not permitted.');
+        }
+
+        return parent::setFlags($flags);
```
Ocramius, line 220: Hmm, wait, the parent class also had flags? :O

test/Generator/PropertyGeneratorTest.php
```diff
@@ -330,4 +380,20 @@ public function testFromReflectionOmitsTypeHintInTypedProperty(): void
 
         self::assertSame('    private $typedProperty;', $code);
     }
+
+    public function testFromReflectionReadonlyProperty(): void
+    {
+        if (false === strpos((string)phpversion(), '8.1')) {
```
Ocramius: This can be `/** @requires php>=8.1 */` at docblock level :+1:

test/Generator/PropertyGeneratorTest.php
```diff
@@ -330,4 +380,20 @@ public function testFromReflectionOmitsTypeHintInTypedProperty(): void
...
+
+            return;
+        }
+
+        $reflectionProperty = new PropertyReflection(ClassWithReadonlyProperty::class, 'readonly');
+
+        $generator = PropertyGenerator::fromReflection($reflectionProperty);
+        $code      = $generator->generate();
+
+        self::assertSame('    private $typedProperty;', $code);
```
Ocramius: This assertion seems to indicate that the `readonly` flag was not added? Seems wrong

Ocramius (changes requested):

src/Generator/PropertyGenerator.php
michaelpetri: This is just an override of the parent in order to throw exceptions for impossible flag combinations

src/Generator/PropertyGenerator.php
michaelpetri: Added types all over in all files i touched.

src/Generator/PropertyGenerator.php
```diff
@@ -112,18 +121,21 @@ public static function fromArray(array $array)
...
         }
 
         return $property;
     }
 
     /**
-     * @param ?string $name
      * @param PropertyValueGenerator|string|array|null $defaultValue
-     * @param int $flags
+     * @param int|int[] $flags
```
michaelpetri, line 147: Because in the tests an int array is used.

src/Generator/AbstractMemberGenerator.php
```diff
@@ -27,9 +27,9 @@ abstract class AbstractMemberGenerator extends AbstractGenerator
 
     /**
      * @param  int|int[] $flags
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setFlags($flags)
+    public function setFlags($flags): self
```
Ocramius: Just realized that we can't add `: self` here, because child classes may be implementing `setFlags()`, and they will break, if they didn't declare `:self`

`:static` at docblock level, as well as type refinements in docblock, are sufficient

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
     }
 
     /**
-     * @param  int $flag
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function addFlag($flag)
+    public function addFlag(int $flag): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
         return $this;
     }
 
     /**
-     * @param  int $flag
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function removeFlag($flag)
+    public function removeFlag(int $flag): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
         return $this;
     }
 
     /**
-     * @param  bool $isAbstract
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setAbstract($isAbstract)
+    public function setAbstract(bool $isAbstract): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
     {
         return $isAbstract ? $this->addFlag(self::FLAG_ABSTRACT) : $this->removeFlag(self::FLAG_ABSTRACT);
     }
 
-    /**
-     * @return bool
-     */
-    public function isAbstract()
+    public function isAbstract(): bool
```
Ocramius: Similar to above: we can't add `:bool`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
         return (bool) ($this->flags & self::FLAG_ABSTRACT);
     }
 
     /**
-     * @param  bool $isInterface
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setInterface($isInterface)
+    public function setInterface(bool $isInterface): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
     {
         return $isInterface ? $this->addFlag(self::FLAG_INTERFACE) : $this->removeFlag(self::FLAG_INTERFACE);
     }
 
-    /**
-     * @return bool
-     */
-    public function isInterface()
+    public function isInterface(): bool
```
Ocramius: Similar to above: we can't add `:bool`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
         return (bool) ($this->flags & self::FLAG_INTERFACE);
     }
 
     /**
-     * @param  bool $isFinal
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setFinal($isFinal)
+    public function setFinal(bool $isFinal): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
     {
         return $isFinal ? $this->addFlag(self::FLAG_FINAL) : $this->removeFlag(self::FLAG_FINAL);
     }
 
-    /**
-     * @return bool
-     */
-    public function isFinal()
+    public function isFinal(): bool
```
Ocramius: Similar to above: we can't add `:bool`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
         return (bool) ($this->flags & self::FLAG_FINAL);
     }
 
     /**
-     * @param  bool $isStatic
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setStatic($isStatic)
+    public function setStatic(bool $isStatic): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
     {
         return $isStatic ? $this->addFlag(self::FLAG_STATIC) : $this->removeFlag(self::FLAG_STATIC);
     }
 
-    /**
-     * @return bool
-     */
-    public function isStatic()
+    public function isStatic(): bool
```
Ocramius: Similar to above: we can't add `:bool`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
     }
 
     /**
-     * @param  string $visibility
-     * @return AbstractMemberGenerator
+     * @psalm-param static::VISIBILITY_* $visibility
+     * @return static
      */
-    public function setVisibility($visibility)
+    public function setVisibility(string $visibility): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -45,98 +45,80 @@ public function setFlags($flags)
...
-    public function isStatic()
+    public function isStatic(): bool
     {
         return (bool) ($this->flags & self::FLAG_STATIC); // is FLAG_STATIC in flags
     }
 
     /**
-     * @param  string $visibility
-     * @return AbstractMemberGenerator
+     * @psalm-param static::VISIBILITY_* $visibility
```
Ocramius: Nice touch: this is some good type improvement :+1:

src/Generator/AbstractMemberGenerator.php
```diff
@@ -157,9 +139,9 @@ public function setVisibility($visibility)
     }
 
     /**
-     * @return string
+     * @psalm-return static::VISIBILITY_*
      */
-    public function getVisibility()
+    public function getVisibility(): string
```
Ocramius: Similar to above: we can't add `:string`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -172,29 +154,25 @@ public function getVisibility()
     }
 
     /**
-     * @param  string $name
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setName($name)
+    public function setName(string $name): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -172,29 +154,25 @@ public function getVisibility()
...
-        $this->name = (string) $name;
+        $this->name = $name;
         return $this;
     }
 
-    /**
-     * @return string
-     */
-    public function getName()
+    public function getName(): string
```
Ocramius: Similar to above: we can't add `:string`

src/Generator/AbstractMemberGenerator.php
```diff
@@ -172,29 +154,25 @@ public function getVisibility()
...
     }
 
     /**
      * @param  DocBlockGenerator|string $docBlock
      * @throws Exception\InvalidArgumentException
-     * @return AbstractMemberGenerator
+     * @return static
      */
-    public function setDocBlock($docBlock)
+    public function setDocBlock($docBlock): self
```
Ocramius: Similar to above: we can't add `:self`

src/Generator/AbstractMemberGenerator.php
Ocramius: Practical example: https://3v4l.org/f2Tu7

src/Generator/PropertyGenerator.php
```diff
@@ -4,13 +4,17 @@
...
+use function method_exists;
+use function PHPUnit\Framework\assertIsBool;
 use function sprintf;
 use function str_replace;
 use function strtolower;
 
 class PropertyGenerator extends AbstractMemberGenerator
 {
     public const FLAG_CONSTANT = 0x08;
+    public const FLAG_READONLY = 0x80;
```
Ocramius, line 20: I wonder if we should set this to `private` :thinking:

src/Generator/PropertyGenerator.php
```diff
@@ -4,13 +4,17 @@
 
 use Laminas\Code\Reflection\PropertyReflection;
 
+use function array_reduce;
+use function method_exists;
+use function PHPUnit\Framework\assertIsBool;
```
Ocramius: This symbol comes from the test suite: we can't use it in sources

src/Generator/PropertyGenerator.php
```diff
@@ -112,18 +121,21 @@ public static function fromArray(array $array)
                 case 'omitdefaultvalue':
                     $property->omitDefaultValue($value);
                     break;
+                case 'readonly':
+                    assertIsBool($value, 'Readonly must be of type bool');
```
Ocramius: `assert(is_bool($value), ...)`

src/Generator/PropertyGenerator.php
```diff
@@ -136,40 +148,62 @@ public function __construct($name = null, $defaultValue = null, $flags = self::F
         }
     }
 
-    /**
-     * @param  bool $const
-     * @return PropertyGenerator
-     */
-    public function setConst($const)
+    public function setConst(bool $const): self
```
Ocramius: `:self` cannot be added here.

`@return static` should probably be kept at docblock level?

src/Generator/PropertyGenerator.php
```diff
@@ -136,40 +148,62 @@ public function __construct($name = null, $defaultValue = null, $flags = self::F
...
-            $this->removeFlag(self::FLAG_CONSTANT);
+        if (true === $const) {
+            return $this->setFlags(self::FLAG_CONSTANT);
         }
 
-        return $this;
+        return $this->removeFlag(self::FLAG_CONSTANT);
+    }
+
+    public function isConst(): bool
```
Ocramius: ~~I would avoid expanding the public API here - let's get rid of this new method~~

EDIT: I see that this method is pre-existing - all good here, but we can't add a native `:bool` to it

src/Generator/PropertyGenerator.php
```diff
@@ -136,40 +148,62 @@ public function __construct($name = null, $defaultValue = null, $flags = self::F
...
+    {
+        return (bool) ($this->flags & self::FLAG_CONSTANT);
+    }
+
+    public function setReadonly(bool $readonly): self
```
Ocramius, line 187: We should probably document `/** @return static */`

src/Generator/PropertyGenerator.php
```diff
@@ -136,40 +148,62 @@ public function __construct($name = null, $defaultValue = null, $flags = self::F
...
+    {
+        if (true === $readonly) {
+            return $this->setFlags(self::FLAG_READONLY);
+        }
+
+        return $this->removeFlag(self::FLAG_READONLY);
+    }
+
+    public function isReadonly(): bool
```
Ocramius, line 198: We should probably get rid of this public API - let's consider `#generate()` our main API, while reading state is probably not relevant (and not the use-case for this class)

src/Generator/PropertyGenerator.php
```diff
@@ -136,40 +148,62 @@ public function __construct($name = null, $defaultValue = null, $flags = self::F
...
+    {
+        return (bool) ($this->flags & self::FLAG_READONLY);
     }
 
     /**
-     * @return bool
+     * {@inheritDoc}
      */
-    public function isConst()
+    public function setFlags($flags): self
```
Ocramius: `: self` cannot be added here

src/Generator/PropertyGenerator.php
```diff
@@ -179,19 +213,15 @@ public function setDefaultValue(
         return $this;
     }
 
-    /**
-     * @return ?PropertyValueGenerator
-     */
-    public function getDefaultValue()
+    public function getDefaultValue(): ?PropertyValueGenerator
```
Ocramius: We can't add the native return type declaration here

src/Generator/PropertyGenerator.php
```diff
@@ -179,19 +213,15 @@ public function setDefaultValue(
...
     {
         return $this->defaultValue;
     }
 
     /**
      * @throws Exception\RuntimeException
-     * @return string
      */
-    public function generate()
+    public function generate(): string
```
Ocramius: We can't add the native return type declaration here

src/Generator/PropertyGenerator.php
```diff
@@ -179,19 +213,15 @@ public function setDefaultValue(
...
-     */
-    public function getDefaultValue()
+    public function getDefaultValue(): ?PropertyValueGenerator
     {
         return $this->defaultValue;
     }
 
     /**
      * @throws Exception\RuntimeException
-     * @return string
```
Ocramius, line 192: One thing we can add is `/** @psalm-return non-empty-string */`

Ocramius (changes requested):

src/Generator/PropertyGenerator.php
michaelpetri: Totally, changed!

src/Generator/PropertyGenerator.php
michaelpetri: :sweat_smile: 
Removed and replaced by condition and exception.

src/Generator/PropertyGenerator.php
michaelpetri: Reverted since we refer to it from tests.

test/Generator/PropertyGeneratorTest.php
michaelpetri: Syntax is `/** @requires PHP >= 8.1 */`

src/Generator/PropertyGenerator.php
michaelpetri: Used condition with `Laminas\Code\Generator\Exception\InvalidArgumentException`, also okay?

src/Generator/PropertyGenerator.php
michaelpetri: Reverted all signature changes but kept `@return static`

src/Generator/PropertyGenerator.php
michaelpetri: Reverted and added psalm non-empty-string

src/Generator/PropertyGenerator.php
michaelpetri: We use it in tests, also this version provides such an interface for all other modifiers, imo we should keep it. Also keep in mind this is just a bool, so not much which can be misused?! But your decision!

Ocramius:
Static analysis failing due to parsing: I think I need to set up dependabot here 🤔

michaelpetri:
> Static analysis failing due to parsing: I think I need to set up dependabot here thinking

I assume the problem here is that psalm runs on php80 and can not parse the test assets which contain php81 syntax?
Maybe the files just need to be excluded. Will check later when we have our half day for research.

michaelpetri:
> > Static analysis failing due to parsing: I think I need to set up dependabot here thinking
> 
> I assume the problem here is that psalm runs on php80 and can not parse the test assets which contain php81 syntax? Maybe the files just need to be excluded. Will check later when we have our half day for research.

The problem is indeed he PHP version `<8.1.0`, unfortunately excluding the file via psalm.xml is not working. As discussed in slack i dropped the Class from code base and replaced it by an `eval()` to generate it at runtime.

Ocramius (approved):
LGTM :ship:

Ocramius:
HMM, static analysis still failed :thinking: 

```
Run vendor/bin/psalm  --output-format=github --shepherd --stats
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  60 / 185 (32%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 120 / 185 (64%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 180 / 185 (97%)
░░░░░
Error: Syntax error, unexpected T_STRING, expecting T_VARIABLE on line 9 (see https://psalm.dev/173)
🐑 results sent to shepherd.dev 🐑
Error: Process completed with exit code 2.
```

michaelpetri:
> HMM, static analysis still failed :thinking: 
> 
> ```
> Run vendor/bin/psalm  --output-format=github --shepherd --stats
> Scanning files...
> Analyzing files...
> 
> ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  60 / 185 (32%)
> ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 120 / 185 (64%)
> ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 180 / 185 (97%)
> ░░░░░
> Error: Syntax error, unexpected T_STRING, expecting T_VARIABLE on line 9 (see https://psalm.dev/173)
> 🐑 results sent to shepherd.dev 🐑
> Error: Process completed with exit code 2.
> ```

Okay that's strange, will try to reproduce local.

michaelpetri:
Seems like i removed my own commit after the soft reset / rebase issues i faced the last time :man_shrugging: 

Anyway, removed class again, added eval to generate it while runtime. Unfortunately i was not able to solve this issue:
```
ERROR: MixedAssignment - test/Generator/PropertyGeneratorTest.php:399:9 - Unable to determine the type that $generator is being assigned to (see https://psalm.dev/032)
        $generator = PropertyGenerator::fromReflection($reflectionProperty);
```
So i decided to add it to baseline, any ideas? 

I also realized that psalm is only running on PHP 8.0 in CI, should we extend matrix here as well?

src/Generator/PropertyGenerator.php
Ocramius: Remember that any API we add is API we have to maintain long-term too.

No big issue here though, let's add it 👍

Ocramius (approved):
LGTM 🚢 

Thanks @michaelpetri!

Issue #104, opened 2021-09-14
PHP 8.1 `new` in initializers support


Ocramius:
https://wiki.php.net/rfc/new_in_initializers

michaelpetri:
This one should be implemented as a new flag as well? `PropertyGenerator::FLAG_IN_INITIALIZER` ?

Ocramius:
Difficult to say: I think this is part of `ParameterGenerator`, not `PropertyGenerator`? :thinking: 

The bulk of all logic lives in the `ValueGenerator`, which is not currently capable of complex expressions:

https://github.com/laminas/laminas-code/blob/e9e2dcc47159900b4187b585aaf3fd8380dc8946/src/Generator/ValueGenerator.php#L27

It may perhaps be something relevant post-release, but the isomorphism between `ParameterGenerator` and `ReflectionParameter` is something we should try to achieve

michaelpetri:
Will give it a try on wednesday @lyska or after my other two open patches :v:

Ocramius:
As discussed with @michaelpetri, the problem with `new` in initializers is that we can't deduct a PHP expression for generating a value, when given a value.

The `ValueGenerator` would need radical adjustments here, but it would not be capable of generating an expression in a reliable way.

For example:

```php
class MyDefaultValue {
    public int $value;
    public function __construct() {
        $this->value = random_int(1, 10000);
    }
}

$value = new MyDefaultGenerator();

// this is potentially wrong: `'new MyDefaultValue()'` is probably not reliable either?
$parameterGenerator->setDefaultValue(ValueGenerator::fromObject($value)); 
```

I don't think it is realistically feasible to solve this problem by `4.5.0`, and therefore removing the milestone.

nicolas-grekas:
This might be doable by casting reflectors to string since https://github.com/php/php-src/pull/7540

PR #114, opened 2021-11-20
Added support for PHP 8.0 promoted parameters

michaelpetri:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

Adds support for promoted parameters

Solves (partially) https://github.com/laminas/laminas-code/issues/104

### DIFF
--- a/src/Generator/ClassGenerator.php
+++ b/src/Generator/ClassGenerator.php
@@ -33,6 +33,7 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
     public const IMPLEMENTS_KEYWORD = 'implements';
     public const FLAG_ABSTRACT      = 0x01;
     public const FLAG_FINAL         = 0x02;
+    private const CONSTRUCTOR_NAME  = '__construct';
 
     protected ?FileGenerator $containingFileGenerator = null;
 
@@ -140,7 +141,17 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
             }
 
             if ($reflectionMethod->getDeclaringClass()->getName() == $className) {
-                $methods[] = MethodGenerator::fromReflection($reflectionMethod);
+                $method = MethodGenerator::fromReflection($reflectionMethod);
+
+                if (self::CONSTRUCTOR_NAME === strtolower($method->getName())) {
+                    foreach ($method->getParameters() as $parameter) {
+                        if ($parameter instanceof PromotedParameterGenerator) {
+                            $cg->removeProperty($parameter->getName());
+                        }
+                    }
+                }
+
+                $methods[] = $method;
             }
         }
 
@@ -855,6 +866,16 @@ class ClassGenerator extends AbstractGenerator implements TraitUsageInterface
                 'A method by name %s already exists in this class.',
                 $methodName
             ));
+        }
+
+        if (self::CONSTRUCTOR_NAME !== strtolower($methodName)) {
+            foreach ($method->getParameters() as $parameter) {
+                if ($parameter instanceof PromotedParameterGenerator) {
+                    throw new Exception\InvalidArgumentException(
+                        'Promoted parameter can only be added to constructor.'
+                    );
+                }
+            }
         }
 
         $this->methods[strtolower($methodName)] = $method;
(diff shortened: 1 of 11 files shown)


test/Generator/ClassGeneratorTest.php
```diff
@@ -1313,4 +1321,98 @@ public function testCorrectImplementNames()
...
+        eval('namespace ' . __NAMESPACE__ . '; class ' . $className . '{ public function __construct( public string \$inInitializer){} }');
+
+        $classGenerator = ClassGenerator::fromReflection(new ReflectionClass(__NAMESPACE__ . '\\' . $className));
+
+        $expectedOutput = <<<EOS
+class ClassWithInInitializer
+{
+    public function __construct(
+        public string $inInitializer
+    );
```
Ocramius: `{}` still needed

test/Generator/ClassGeneratorTest.php
```diff
@@ -1313,4 +1321,98 @@ public function testCorrectImplementNames()
...
+                )
+            )
+        ]);
+
+        $expectedOutput = <<<EOS
+class InInitializerWithDefaultValue
+{
+    public function __construct(
+        private Foo \$inInitializer = Foo::bar(baz: 'qux')
+    );
```
Ocramius: `{}` still needed

test/Generator/ClassGeneratorTest.php
```diff
@@ -1313,4 +1321,98 @@ public function testCorrectImplementNames()
...
+        $classGenerator->addMethod('__construct', [
+            new InInitializerGenerator('inInitializer', 'string')
+        ]);
+
+        $expectedOutput = <<<EOS
+class InInitializer
+{
+    public function __construct(
+        public string \$inInitializer
+    );
```
Ocramius: `{}` still needed

test/Generator/ClassGeneratorTest.php
```diff
@@ -1313,4 +1321,98 @@ public function testCorrectImplementNames()
...
+
+        self::assertEquals($expectedOutput, $classGenerator->generate());
+    }
+
+    /** @requires PHP >= 8.1 */
+    public function testClassFromReflectionWithInInitializer(): void
+    {
+        $className = uniqid('ClassWithInInitializer', false);
+
+        eval('namespace ' . __NAMESPACE__ . '; class ' . $className . '{ public function __construct( public string \$inInitializer){} }');
```
Ocramius: `\` not needed here

src/Generator/ObjectGenerator.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Code\Generator;
+
+final class ObjectGenerator
```
Ocramius: Discussed in private call: we will not implement this for now

Ocramius: We'll need a new issue for this - will create one after clearing up ideas about this, for now, and for now, we will raise an exception in `::fromReflection()`, when an `object` as default value is encountered (something like "unsupported, for now")

Ocramius (changes requested):
Discussed this review in a 1-on-1 session: PHP 8.1 is currently a bit out of reach.


```php
<?php

interface Bar {}
class Baz implements Bar {}
class Foo
{
    public function __construct(
        public Bar $bar = new Baz()
    ) {}
    
    public function doThing(Bar $bar = new Baz()) {
        return $bar;
    }
}

var_dump(new Foo());

var_dump((new Foo)->doThing());
```

https://3v4l.org/SoOPU

In practice, we should stop this patch at only supporting initializers (and visibility thereof), while we will skip `new_in_initializers` for now ( https://wiki.php.net/rfc/new_in_initializers ) because reflection for it does not give us enough information about the default value yet.

We will likely need AST-based reflector for that, and we're far from that: at that point, this library will either need phasing out, or a full rewrite.

src/Generator/ClassGenerator.php
```diff
@@ -140,7 +141,17 @@ public static function fromReflection(ClassReflection $classReflection)
             }
 
             if ($reflectionMethod->getDeclaringClass()->getName() == $className) {
-                $methods[] = MethodGenerator::fromReflection($reflectionMethod);
+                $method = MethodGenerator::fromReflection($reflectionMethod);
+
+                if (self::CONSTRUCTOR_NAME === $method->getName()) {
```
michaelpetri: Should probably better moved to `MethodGenerator::isConstructor()`

Ocramius: Would be fine to keep it as-is for now, IMO. Just `strtolower()` needed

michaelpetri: Added

Ocramius (approved):
LGTM, excellent work, @michaelpetri!

PR #107, opened 2021-09-23
MethodGenerator::fromArray() add config keys

WinterSilence:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Add keys `'returntype'/setReturnType()` and  `'returnsreference'/setReturnsReference()` to array passed in `MethodGenerator::fromArray()`.

PR #119, opened 2021-12-06
MethodGenerator::fromArray() add config keys

Brammm:
Continuation of https://github.com/laminas/laminas-code/pull/107

### DIFF
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -115,8 +115,8 @@ class MethodGenerator extends AbstractMemberGenerator
      * Generate from array
      *
      * @configkey name             string        [required] Class Name
-     * @configkey docblock         string        The docblock information
-     * @configkey flags            int           Flags, one of MethodGenerator::FLAG_ABSTRACT MethodGenerator::FLAG_FINAL
+     * @configkey docblock         string        The DocBlock information
+     * @configkey flags            int           Flags, one of self::FLAG_ABSTRACT, self::FLAG_FINAL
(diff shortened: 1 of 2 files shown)


test/Generator/MethodGeneratorTest.php
```diff
@@ -307,6 +307,37 @@ public function testCreateFromArray()
...
+        $methodGenerator = MethodGenerator::fromArray([
+            'name'             => 'SampleMethod',
+            'returnsreference' => $value,
+        ]);
+
+        self::assertSame($expected, $methodGenerator->isReturnsReference());
+    }
+
+    /**
+     * @return string[][]
```
Brammm, line 326: Just realized this type is wrong (copied from somewhere else and then forgotten about), though not sure about conventions in the project as to what I should change it to.

src/Generator/MethodGenerator.php
```diff
@@ -310,6 +314,11 @@ public function setReturnsReference($returnsReference)
         return $this;
     }
 
+    public function isReturnsReference(): bool
```
Ocramius, line 317: The `is` should really be removed here - naming is totally wonky (will rename on merge)

test/Generator/MethodGeneratorTest.php
Ocramius: I'll adjust this on merge :+1:

Ocramius (approved):

Ocramius:
Merged via #121, thanks @Brammm!

PR #128, opened 2022-02-21
Prepare for Renovate with reusable workflows

ghostwriter:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Adding Renovate to Laminas, Mezzio repositories. https://gist.github.com/weierophinney/49ca8ae2b916d898dd3d4162b26ff41b

- Configure minimum supported PHP version
- Update composer dependencies
- Update to reusable `Continuous Integration` workflow
- Update to reusable `Automatic Releases` workflow

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <psalm
-        totallyTyped="true"
+        errorLevel="1"
+        phpVersion="8.1"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="https://getpsalm.org/schema/config"
         xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,11 @@
+name: "Continuous Integration"
+
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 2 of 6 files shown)


Ocramius (changes requested):
Requires rolling back to a much smaller `composer.lock` change (just hash change)

Ocramius:
Hmm, the psalm issues seem to be unrelated: I think the reason for the failures is that `vimeo/psalm` started used `config.platform.php` to determine which version it is running on.

We need to add `phpVersion=""` to `psalm.xml` here, I think.

psalm.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <psalm
         errorLevel="1"
+        phpVersion="7.4"
```
Ocramius: Sorry, wasn't clear - the PHP version should probably be 8.1, which is where psalm is run :)

Ocramius (changes requested):

psalm.xml
ghostwriter: I understood that part but got confused because we are still running `ci` check on the minimum supported PHP version,`PHP 7.4`.

Ocramius: Yeah, I'm generally confused by the CI output anyway :shrug: 

Probably need to get back to this at a later point in time, checking it out locally, because those CI failures should really not be there, unless the PHP version changed some details internally too :thinking:

PR #132, opened 2022-06-06
`ReflectionEnum` is no longer `final` in newer PHP 8.1.x releases 

bendavies:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

[ReflectionEnum](https://www.php.net/manual/en/class.reflectionenum.php) is not final.
this somehow causes errror when psalm and phpstan are installed together, and results in:
```
PHP Fatal error:  Class PHPStan\BetterReflection\Reflection\Adapter\ReflectionEnum may not inherit from final class (ReflectionEnum) in phar:///.../vendor/phpstan/phpstan/phpstan.phar/vendor/ondrejmirtes/better-reflection/src/Reflection/Adapter/ReflectionEnum.php on line 30
```

### DIFF
--- a/polyfill/ReflectionEnumPolyfill.php
+++ b/polyfill/ReflectionEnumPolyfill.php
@@ -5,7 +5,7 @@ if (\PHP_VERSION_ID >= 80100) {
 }
 
 /** @internal */
-final class ReflectionEnum
+class ReflectionEnum
 {
     public function __construct($enum)
     {


Ocramius (approved):
Thanks @bendavies!

bendavies:
thankyou!

PR #143, opened 2022-08-04
Value generator enum support

Rastusik:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

This patch enables support for enumerations in ValueGenerator, which is currently missing. It is needed to properly generate proxy classes which have enums as default values in method arguments.

### DIFF
--- a/test/Generator/TestAsset/TestEnum.php
+++ b/test/Generator/TestAsset/TestEnum.php
@@ -1,0 +1,9 @@
+<?php
+
+namespace LaminasTest\Code\Generator\TestAsset;
+
+enum TestEnum
+{
+    case Test1;
+    case Test2;
+}
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "7.4"
+          - "8.0"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -17,7 +17,6 @@ jobs:
           - "highest"
           - "locked"
         php-version:
-          - "7.4"
           - "8.0"
           - "8.1"
         operating-system:
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": ">=7.4, <8.2"
+        "php": ">=8.0, <8.2"
     },
     "require-dev": {
         "ext-phar": "*",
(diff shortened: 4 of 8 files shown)


Rastusik:
ok, I have no idea why the static analysis check doesn't work, but it seems like an error in psalm

Ocramius:
Beware that static analysis runs on PHP 7.4.

Ocramius:
BTW, dropping PHP 7.4 support is also pretty much OK.

Rastusik:
@Ocramius what are you suggesting for me to do? I'm not sure what exactly you mean. should I do the 7.4 support drop? I know that the check runs on 7.4, but the code works even on 7.4 so I'm not sure what to do right now.

Ocramius:
Yeah, try removing 7.4 from `composer.json` (and update the `config.platform`)

Rastusik:
ok so I did what I could but psalm doesn't seem to work correctly

test/Generator/TestAsset/TestEnum.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace LaminasTest\Code\Generator\TestAsset;
+
+enum TestEnum
```
Ocramius, line 5: Ah, enums were introduced in 8.1, hence the parser failure :thinking:

Ocramius:
I think the patch is sound, but in order to raise us to enum support, we'd have to drop PHP 8.1 too.

Psalm correctly infers that the codebase should be verified with PHP 8.0 (compatible):

```
Run vendor/bin/psalm  --output-format=github --shepherd --stats
Target PHP version: 8.0 (inferred from composer.json)
```

Unsure what to do here, will need to think about it.

Rastusik:
yes, that is why I didn't want to do anything more proactively. I'll be waiting for feedback.

Rastusik:
> 
> Unsure what to do here, will need to think about it.

hey @Ocramius , how about this patch? Is there any ETA?

Ocramius:
No ETA - haven't worked on this yet.

test/Generator/TestAsset/TestEnum.php
Ocramius: Ok, yeah, I'm gonna pull PHP 8.0 support, but gonna work on the other issues in the milestone first.

Ocramius:
Couldn't push to your fork, so I created #166 instead.

Note how I fixed a bad bug (TM) where the generated value wasn't considering the current namespace :D

Ocramius:
Thanks @Rastusik!

Rastusik:
> Note how I fixed a bad bug (TM) where the generated value wasn't considering the current namespace :D

nice :D 

thanks for merging @Ocramius

Issue #142, opened 2022-08-03
PHP 8.2 compability


salerat:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary
Due installing by composer i get next error while using PHP 8.2 RC:
`- laminas/laminas-code 4.7.x-dev requires php >=7.4, <8.2 -> your php version (8.2.0beta1) does not satisfy that requirement.`
Looks like this is last package which includes in Symfony 6.2 which not supports PHP 8.2

Ocramius:
Correct: this package does not yet support PHP 8.2.

PHP 8.2 compatibility will be rolled out once all features from PHP 8.2 are implemented/tested:

 * https://wiki.php.net/rfc/dnf_types
 * https://wiki.php.net/rfc/true-type
 * https://wiki.php.net/rfc/readonly_classes
 * https://wiki.php.net/rfc/null-false-standalone-types

agustingomes:
Now that the PHP 8.2 has reached Release Candidate phase, I think is a good moment to add the new features, and in the process, drop support for PHP 7.4.

I'll personally try to invest some hours in the upcoming weeks to get some of the new 8.2 features in.

guradia:
I have a codebase with Symfony 6.1.x at hand, that does work fine under PHP 8.2.
But doing anything with composer gets blocked by this constraint.

So i think the constraint is kind of erroneously applied?

Meaning:
Coming from the view of features this library understands, sure ... it does not support PHP 8.2.
But the constraints applied to a library should be about the compatibility of the library itself.
So to speak: PHP 8.2 understands the library and can run it's things ... the other way around should not be relevant and only be something the documentation of the library states, but does not dictate it can't be installed under higher PHP-Versions.

Addendum:
My understanding of this library is, that it's compatibility is meant in way of what it can statically analyse. And this is separate from it's one execution context. It can probably analyse 8.1 code, even when executed on 8.0?

Ocramius:
@guradia just cross-referencing: https://github.com/laminas/laminas-diactoros/issues/117#issuecomment-1267306142

As for PHP 8.2 support, hoping to throw some time at this soon: haven't found any yet, and I've already upgraded a few hundred other packages so far. Patience.

guradia:
@Ocramius ah, yes. This reasoning about SemVer is totally sound. And to be square with myself: Me going early adopter is obviously calling for trouble and is my problem alone.

[...]

PR #149, opened 2022-11-02
Add support for standalone `true`, `false` and `null` types

IonBazan:
Signed-off-by: Ion Bazan <ion.bazan@gmail.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Provides support for `true`, `false` and `null` types. 

The `false` and `null` ale no longer required to be a part of a union, therefore they can used on its own. The `requiresUnionWithStandaloneType()` method has been replaced with a simple `return false` as it is used in `assertCanBeAStandaloneType()` and `assertCanIntersectWith()` - it may be replaced in the future if needed.

See: #142

### DIFF
--- a/src/Generator/TypeGenerator/AtomicType.php
+++ b/src/Generator/TypeGenerator/AtomicType.php
@@ -42,14 +42,16 @@ final class AtomicType
         'mixed'    => 10,
         'void'     => 11,
         'false'    => 12,
-        'null'     => 13,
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Thanks @IonBazan!

PR #150, opened 2022-11-04
fix: complete ReflectionEnum polyfill

alanpoulain:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Since the polyfill is autoloaded, PHPStan seems to discover the class, even with the early return:
```
Call to an undefined method ReflectionEnum::getCases()
```

Completing this polyfill solves the issue.

### DIFF
--- a/polyfill/ReflectionEnumPolyfill.php
+++ b/polyfill/ReflectionEnumPolyfill.php
@@ -5,10 +5,27 @@ if (\PHP_VERSION_ID >= 80100) {
 }
 
 /** @internal */
-class ReflectionEnum
+class ReflectionEnum extends \ReflectionClass
 {
-    public function __construct($enum)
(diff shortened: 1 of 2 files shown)


polyfill/ReflectionEnumPolyfill.php
```diff
@@ -5,10 +5,27 @@
 }
 
 /** @internal */
-class ReflectionEnum
+class ReflectionEnum extends \ReflectionClass
```
Ocramius, line 8: I decided to drop PHP 8.0 support anyway, which will remove the polyfill entirely.

That said, patch stays open until I'm done with my changes.

Ocramius:
Handled in #154

PR #158, opened 2022-11-21
Use `str_starts_with()` and `str_contains()` where possible

fezfez:


Ocramius:
```
Error: src/Generator/ValueGenerator.php:315:46: MixedArgument: Argument 2 of str_contains cannot be mixed, expecting string (see https://psalm.dev/030)
```

Unsure if it should go to baseline or not.

src/Generator/ValueGenerator.php
```diff
@@ -311,7 +312,7 @@ public function getAutoDeterminedType($value)
                         return self::TYPE_CONSTANT;
                     }
 
-                    if (strpos($value, $constant) !== false) {
+                    if (str_contains($value, $constant)) {
```
Ocramius, line 315: Yeah, probably to baseline...

fezfez: i'he try to find a solution but : 

```
ERROR: TooManyTemplateParams - src/Generator/ValueGenerator.php:70:13 - Laminas\Stdlib\ArrayObject<array-key, string> has too many template params, expecting 0 (see https://psalm.dev/184)
     * @var SplArrayObject<array-key, string>|StdlibArrayObject<array-key, string>
```

Ocramius: I think a suppression is OK here: no need to add further complexity in internal types :thinking:

fezfez: done

Ocramius (approved):
Thanks @fezfez!

PR #159, opened 2022-11-21
Add support for DNF types

IonBazan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no (not yet)
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Basic support for DNF types (as per #142)

Code review welcome, I'd like to know if this is the right direction. Collaboration is welcome too 🙏🏻 .


src/Generator/TypeGenerator/AtomicType.php
```diff
@@ -128,6 +128,11 @@ public function fullyQualifiedName(): string
             : '\\' . $this->type;
     }
 
+    public function __toString(): string
+    {
+        return $this->fullyQualifiedName();
+    }
```
Ocramius: Let's use `->toString()` please :) 

Rationale: https://github.com/ShittySoft/symfony-live-berlin-2018-doctrine-tutorial/pull/3#issuecomment-460441229

src/Generator/TypeGenerator/BaseComplexType.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Laminas\Code\Generator\TypeGenerator;
+
+abstract class BaseComplexType implements Type
```
Ocramius: `/** @internal */` for anything we didn't build explicitly for external systems

Ocramius: s/`BaseComplexType`/`CompositeType`? 

We already know it's a "base" type from the `abstract` keyword

src/Generator/TypeGenerator/BaseComplexType.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * @param AtomicType[]|IntersectionType[] $types
+     */
+    final protected function __construct(protected array $types)
```
Ocramius: Uhh, let's rather please just use union types, instead of an abstract class with `protected` fields.

Really messy to maintain this later on otherwise :)

test/Generator/TypeGenerator/UnionTypeTest.php
```diff
@@ -0,0 +1,27 @@
...
+        $type = UnionType::fromString($typeString);
+
+        self::assertSame($expectedReturnType, $type->__toString());
+    }
+
+    public static function validType(): iterable
+    {
+        yield ['A|B', '\\A|\\B'];
+        yield ['foo|bar|baz|null', '\\foo|\\bar|\\baz|null'];
+        yield ['(foo&bar)|baz|null', '(\\foo&\\bar)|\\baz|null'];
```
Ocramius: We need `(foo&bar)|(baz&taz)|(tar&war&waz)` or such to show that unions are not just across 2 types

Ocramius (changes requested):
Let's get psalm happy, before continuing this :D

IonBazan:
The tests are passing on both PHP 8.1 and 8.2. I added some more test cases too.

Not sure if you want to keep `CompositeTypeTest` as it overlaps with what `TypeGeneratorTest` checks but it's definitely much easier to debug things having a separate test.

There is a lot of mess and code duplication in `toString()` and `__toString()` methods - not quite sure how to handle that nicely but I'm open to suggestions. Technically I could move everything back to the `TypeGenerator` and allow storing `TypeGenerator[]` in `types` but I feel it would give it too many responsibilities. Besides, it doesn't have a common interface with `AtomicType` (and shouldn't).

Idk why is psalm crashing 🤷🏻‍♂️

src/Generator/TypeGenerator/TypeInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Laminas\Code\Generator\TypeGenerator;
+
+use Stringable;
+
+interface TypeInterface extends Stringable
```
Ocramius: Let's keep this stuff `internal` too :D

Ocramius (commented):
Overall looking good from the tests: implementation will need a lot of careful reading / focus :) 

Otherwise nice work! 

I'll check it out locally when I can, and get it shipped.

Ocramius:
The crash in psalm is... novel :O 

I wonder if we should just try going `vimeo/psalm:^5.0.0-rc1`?

IonBazan:
Psalm v5 crashes there too so no solution for that so far 😝

src/Generator/TypeGenerator.php
```diff
@@ -136,30 +129,14 @@ public static function fromTypeString(string $type): self
...
-     */
-    private function __construct(array $types, bool $nullable, bool $isIntersection)
+    private function __construct(private readonly TypeInterface $type, private readonly bool $nullable = false)
     {
-        $this->types          = $types;
-        $this->nullable       = $nullable;
-        $this->isIntersection = $isIntersection;
+        if ($nullable && $type instanceof AtomicType) {
+            $type->assertCanBeStandaloneNullable();
+        }
```
IonBazan: Should I keep it here or better move back to `fromString()`?

IonBazan:
Well, Psalm is kinda fixed now - produces many more errors but at least the analysis completes. You may want to add them to baseline later.
image

Locked at `dev-master` as `v5.0.0-rc1` does not allow PHP 8.2.

Ocramius:
@IonBazan I'm picking this up: `vimeo/psalm:^5.1.0` fixed the issue with scanning DNF types, apparently :)

IonBazan:
Great, thanks! Let me know if any help is needed.

src/Generator/TypeGenerator.php
Ocramius: Works here too :+1:

src/Generator/TypeGenerator/AtomicType.php
```diff
@@ -128,6 +128,11 @@ public function fullyQualifiedName(): string
             : '\\' . $this->type;
     }
 
+    public function __toString(): string
+    {
+        return $this->type;
+    }
```
Ocramius: `non-empty-string` needed

src/Generator/TypeGenerator/CompositeType.php
```diff
@@ -0,0 +1,157 @@
...
+
+                return $type instanceof self && $type->isIntersection() ? sprintf('(%s)', $typeString) : $typeString;
+            },
+            $this->types
+        );
+
+        return implode($this->getSeparator(), $typesAsStrings);
+    }
+
+    public function fullyQualifiedName(): string
```
Ocramius: Should be `non-empty-string`

src/Generator/TypeGenerator/CompositeType.php
```diff
@@ -0,0 +1,157 @@
...
+        return $this->isIntersection;
+    }
+
+    /** @return self::INTERSECTION_SEPARATOR|self::UNION_SEPARATOR */
+    public function getSeparator(): string
+    {
+        return $this->isIntersection ? self::INTERSECTION_SEPARATOR : self::UNION_SEPARATOR;
+    }
+
+    public function __toString(): string
```
Ocramius: Should be `non-empty-string`

src/Generator/TypeGenerator/CompositeType.php
```diff
@@ -0,0 +1,157 @@
...
+ * This means that this object can be composed of {@see AtomicType} or other {@see CompositeType} objects.
+ *
+ * @internal the {@see CompositeType} is an implementation detail of the type generator,
+ *
+ * @psalm-immutable
+ */
+final class CompositeType implements TypeInterface
+{
+    public const UNION_SEPARATOR        = '|';
+    public const INTERSECTION_SEPARATOR = '&';
```
Ocramius: Representing both union and intersection types as a single type is quite a mess: IMO not a solution we should roll with.

Ocramius (commented):
Still working on it - got stuck on paid work :P

Ocramius:
@IonBazan as you can see, I ended up using a radically different representation of internal types, mostly so that it can be understood at runtime.

Shipping this now :)

PR #154, opened 2022-11-19
Drop PHP 7.4 and PHP 8.0 support

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

This patch overhauls the various types in use in this package, and requires PHP 8.1 to do so, and to achieve parity with the current ENUM integrations.

TODOs:

 * [x] make the build green (yes, I **did** break something)
 * [x] adjust CI so that it no longer runs PHP 8.0

Improvement:

```diff
-Psalm was able to infer types for 95.8537% of the codebase
+Psalm was able to infer types for 96.5310% of the codebase
```

PR #161, opened 2022-11-25
Make the return-type of `ClassReflection#getStartLine()` explicit

nicolas-grekas:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

This was removed in #154 but this phpdoc is useful to some static analyzers, so that they can know what the future return-type is going to be.

### DIFF
--- a/src/Reflection/ClassReflection.php
+++ b/src/Reflection/ClassReflection.php
@@ -44,7 +44,8 @@ class ClassReflection extends ReflectionClass implements ReflectionInterface
     /**
      * {@inheritDoc}
      *
-     * @param bool $includeDocComment
+     * @param  bool $includeDocComment
+     * @return int|false
      */
     #[ReturnTypeWillChange]
     public function getStartLine($includeDocComment = false)


src/Reflection/ClassReflection.php
```diff
@@ -44,7 +44,8 @@ public function getDocBlock()
     /**
      * {@inheritDoc}
      *
-     * @param bool $includeDocComment
+     * @param  bool $includeDocComment
+     * @return int|false
```
Ocramius, line 48: I would suggest trying `@inheritDoc` (no brackets) first.

Overall not opposed to this change, but to be attempted first.

Ocramius (changes requested):

src/Reflection/ClassReflection.php
Ocramius: Wait... is this again because of Symfony's debug loader?

IonBazan:
If we are aiming for `~8.1.0 || ~8.2.0` for v4.8.0, I think we don't really need to use `#[ReturnTypeWillChange]` and replace it with native union type there, right?

src/Reflection/ClassReflection.php
nicolas-grekas: There's no `@inheritDoc` data to read on the parent (internal) class unfortunately so this doesn't provide the needed info.

Ocramius:
@IonBazan changing a return type in a non-`final` class is still a BC break

src/Reflection/ClassReflection.php
Ocramius: Fair enough: I still smell the stench of `symfony/debug`, but I'll let this slip :P

IonBazan:
Noted, let's do that for v5, if you don't mind

Ocramius:
@IonBazan I'd be really happy to make a lot of this stuff `final` for v5.

Once it's `final`, it is much much easier to maintain and change internals.

nicolas-grekas:
Thanks for merging quickly!

Ocramius:
No problem: just beware that I won't release until @IonBazan's #159 is also merged.

PR #173, opened 2023-03-03
Non-variadic parameter should not throw exception when calling `ParameterGenerator#setVariadic(false)` with an existing default value

sakhunzai:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Fix the changes introduce in PR https://github.com/laminas/laminas-code/pull/72 by ensuring call to `setVariadic` should not throw exception when parameter is non variadic.

### DIFF
--- a/test/Generator/ParameterGeneratorTest.php
+++ b/test/Generator/ParameterGeneratorTest.php
@@ -558,8 +558,9 @@ class ParameterGeneratorTest extends TestCase
         $parameter->setType('int');
         $parameter->setPosition(1);
         $parameter->setVariadic(false);
-        $this->expectNotToPerformAssertions();
-        $parameter->setDefaultValue([]);
+        self::assertSame('int $parameter', $parameter->generate());
+        $parameter->setDefaultValue(7);
+        self::assertSame('int $parameter = 7', $parameter->generate());
     }
 
     public function testMakingParameterVariadicWithExistingDefaultValueThrowsInvalidArgumentException(): void
@@ -585,9 +586,10 @@ class ParameterGeneratorTest extends TestCase
         $parameter->setName('parameter');
         $parameter->setType('int');
         $parameter->setPosition(1);
-        $parameter->setDefaultValue([]);
-        $this->expectNotToPerformAssertions();
+        $parameter->setDefaultValue(7);
+        self::assertSame('int $parameter = 7', $parameter->generate());
         $parameter->setVariadic(false);
+        self::assertSame('int $parameter = 7', $parameter->generate());
     }
 
     #[Group('zendframework/zend-code#29')]
--- a/test/Generator/ParameterGeneratorTest.php
+++ b/test/Generator/ParameterGeneratorTest.php
@@ -561,7 +561,7 @@ class ParameterGeneratorTest extends TestCase
         $this->expectNotToPerformAssertions();
         $parameter->setDefaultValue([]);
     }
-    
+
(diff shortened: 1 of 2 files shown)


test/Generator/ParameterGeneratorTest.php
```diff
@@ -550,6 +550,18 @@ public function testAssigningDefaultValueToVariadicParameterThrowsInvalidArgumen
...
+    public function testAssigningDefaultValueToNonVariadicParameter(): void
+    {
+        $parameter = new ParameterGenerator();
+
+        $parameter->setName('parameter');
+        $parameter->setType('int');
+        $parameter->setPosition(1);
+        $parameter->setVariadic(false);
+        $this->expectNotToPerformAssertions();
+        $parameter->setDefaultValue([]);
```
Ocramius: This test should verify the generated code

test/Generator/ParameterGeneratorTest.php
```diff
@@ -566,6 +578,18 @@ public function testMakingParameterVariadicWithExistingDefaultValueThrowsInvalid
...
+    public function testMakingParameterNonVariadicWithExistingDefaultValue(): void
+    {
+        $parameter = new ParameterGenerator();
+
+        $parameter->setName('parameter');
+        $parameter->setType('int');
+        $parameter->setPosition(1);
+        $parameter->setDefaultValue([]);
+        $this->expectNotToPerformAssertions();
+        $parameter->setVariadic(false);
```
Ocramius: This test should verify the generated code

Ocramius (changes requested):

test/Generator/ParameterGeneratorTest.php
sakhunzai: There is nothing to verify since there is no assertion for not throwing exceptions.

Ocramius: No, I'd expect to see what the generated code looks like for both tests.

sakhunzai: This is not enough [f3bca9a](https://github.com/laminas/laminas-code/pull/173/commits/f3bca9a1ae6b1c4367ab614e88d622e65313aebe) ?

Ocramius: Nay, please do expand the test with the generated code expectation, and remove the `expectNotToPerformAssertions` call.

sakhunzai: Given this https://github.com/sebastianbergmann/phpunit/issues/2484, I have no idea what we can test here.

sakhunzai: Or do you think sth like this will work:

```
public function testAssertionPassesWhenFooBarIsAvailable()
{
    try {
        $this->fooBarAvailabilityChecker->assertFooBarAvailable();

        $this->addToAssertionCount(1);
    } catch (FooBarNotAvailable $exception) {
        $this->fail($exception->getMessage());
    }
}
```

As suggested in the thread

Ocramius: `self::assertSame('the generated string', $parameter->generate())`

sakhunzai: Ah, thanks I had no idea  :)

sakhunzai:
please review changes

Ocramius (approved):
Excellent, thanks @sakhunzai!

PR #174, opened 2023-03-05
Use output mode configured in `ValueGenerator` instance

thomas-kl1:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

When I want to generate code for an array with the "single line" output, I expect the result to be on a single line. However in the case the array have deeper depth than 1, the result is on multi lines, because the original output mode is not given when generating the child nodes.

### DIFF
--- a/test/Generator/ValueGeneratorTest.php
+++ b/test/Generator/ValueGeneratorTest.php
@@ -19,7 +19,6 @@ use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\TestCase;
 
 use function fopen;
-use function sprintf;
 use function str_replace;
 
 #[CoversClass(ValueGenerator::class)]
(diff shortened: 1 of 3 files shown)


Ocramius (changes requested):
Needs a test

thomas-kl1:
Hi @Ocramius 
I've pushed the test as well. Do you want me to add a suppress warning for the psalm rule "unused method" on the data provider?

test/Generator/ValueGeneratorTest.php
```diff
@@ -498,4 +497,55 @@ public function testExceptionInvalidValue(mixed $value, string $type): void
...
+    ): void {
+        $valueGenerator = new ValueGenerator($array, $type, $outputMode);
+
+        self::assertSame($valueGenerator->generate(), $output);
+    }
+
+    /**
+     * Data provider for testArrayWithOutputMode test
+     */
+    public static function multipleOutputArray(): array
```
Ocramius, line 519: This error is expected: PHPUnit 10 is not yet understood by `psalm/plugin-phpunit`.

Re-generating the baseline should fix this.

thomas-kl1: I'm not sure what to do with this information, can I do something there?

froschdesign: @thomas-kl1 
See: https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#using-a-baseline-file

Ocramius (approved):
Thanks @thomas-kl1!

Issue #182, opened 2023-05-16
Promoted properties missing default value


n-valverde:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 4.11.0 (all versions affected since 4.5.0)

#### Summary

Promoted properties support was added by https://github.com/laminas/laminas-code/pull/114 
For some reason, this is lacking support for promoted property default value, I am not sure if this was made on purpose or not, but adding this seems pretty straight forward

Full context of the issue: In a Symfony application, when configuring a service as lazy, this is generating a proxy under the hood (using friendsofphp/proxy-manager-lts). Hunting all deprecations on the road of upgrading PHP in my userland application, I've been faced with a notice about optional parameter being treated as required because of the parameter position. Problem is, this notice come from the generated proxy, but my userland code signature is correct. Let's consider this signature

```php
class SomeService
{
    private LoggerInterface $logger;

    public function __construct(private SomeDependency $dependency, ?LoggerInterface $logger = null, private string $someOptionalConfig = "")
    {
        $this->logger = $logger ?? new NullLogger();
    }
}
```

That signature is a bit ugly but is technically correct, the problem is, when generating a proxy for this class, you end up with

```php
class SomeService
{
    private LoggerInterface $logger;

    public function __construct(private SomeDependency $dependency, ?LoggerInterface $logger = null, private string $someOptionalConfig)
    {
        $this->logger = $logger ?? new NullLogger();
    }
}
```

`$someOptionalConfig` does not have its default value anymore, so the argument position becomes wrong because there is this optional `$logger` which is now treated as required.

Obviously I can change the userland code signature to fix that, but this behaviour feels unexpected

#### Current behavior

Promoted property default value is basically always ignored

#### How to reproduce

Slighlty changing the relevant test easily show the failure, see below

```php
namespace LaminasTest\Code\Generator\TestAsset;

final class ClassWithPromotedParameter
{
    public function __construct(private string $promotedParameter = 'foobar') {
    }
}
```

```php
namespace LaminasTest\Code\Generator;

class ClassGeneratorTest extends TestCase
{
    public function testClassWithPromotedParameterFromReflection(): void
    {
        $classGenerator = ClassGenerator::fromReflection(
            new ClassReflection(ClassWithPromotedParameter::class)
        );

        $expectedOutput = <<<EOS
namespace LaminasTest\Code\Generator\TestAsset;

final class ClassWithPromotedParameter
{
[...]

PR #184, opened 2023-05-16
Nullable flag on constructor property promotion, closes #183

### LINKED ISSUE #183: Promoted properties are losing nullable flag
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 4.7.1 (all versions affected since 4.5.0)

#### Summary

Laminas\Code\Generator\PromotedParameterGenerator#fromParameterGeneratorWithVisibility are losing nullable flag of such parameter.

#### Current behavior

```php
class Test {
    public function __construct(
        protected ?int $value
    ) {
    }
}

$parameterReflection = new ParameterReflection([Test::class, '__construct'], 0);
$parameterGenerator = PromotedParameterGenerator::fromReflection($parameterReflection);
echo $parameterGenerator->generate();
```

will echo `protected int $value`, nullable flag is lost.

#### How to reproduce

Use snippet above.

#### Expected behavior

Code above should echo  `protected ?int $value`.

#### Additional information

https://github.com/laminas/laminas-code/blob/169123b3ede20a9193480c53de2a8194f8c073ec/src/Generator/PromotedParameterGenerator.php#LL73C51-L73C51
```php
    public static function fromParameterGeneratorWithVisibility(ParameterGenerator $generator, string $visibility): self
    {
        $name = $generator->getName();
        $type = $generator->getType(); // <----- this would return "bare" type, without modifiers
```

Unfortunately, there are no way to get nullable flag from TypeGenerator at all for now. Unsurprisingly it get lost on the way.

Grundik:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

See #183 for details.

### DIFF
--- a/src/Generator/ParameterGenerator.php
+++ b/src/Generator/ParameterGenerator.php
@@ -168,6 +168,14 @@ class ParameterGenerator extends AbstractGenerator
     }
 
     /**
+     * @return bool
+     */
+    public function getNullable()
+    {
+        return $this->type && $this->type->getNullable();
+    }
+
+    /**
      * @param  string $name
      * @return ParameterGenerator
      */
--- a/src/Generator/PromotedParameterGenerator.php
+++ b/src/Generator/PromotedParameterGenerator.php
@@ -87,6 +87,10 @@ final class PromotedParameterGenerator extends ParameterGenerator
             );
         }
 
+        if ($type && $generator->getNullable()) {
+            $type = '?' . $type;
+        }
+
         return new self(
             $name,
             $type,
--- a/src/Generator/TypeGenerator.php
+++ b/src/Generator/TypeGenerator.php
@@ -141,6 +141,14 @@ final class TypeGenerator implements GeneratorInterface
     }
 
     /**
+     * @return bool Nullable flag
+     */
+    public function getNullable()
+    {
+        return $this->nullable;
+    }
+
+    /**
      * @return bool[]|string[] ordered tuple, first key represents whether the type is nullable, second is the
      *                         trimmed string
      * @psalm-return array{bool, string}
--- a/test/Generator/PromotedParameterGeneratorTest.php
+++ b/test/Generator/PromotedParameterGeneratorTest.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\Code\Generator;
+
+use Laminas\Code\Generator\PromotedParameterGenerator;
+use Laminas\Code\Reflection\ParameterReflection;
+use LaminasTest\Code\TestAsset\ClassWithPromotedProperties;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\TestCase;
+
+#[Group('Laminas_Code_Generator')]
+#[Group('Laminas_Code_Generator_Php')]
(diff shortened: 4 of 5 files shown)


src/Generator/ParameterGenerator.php
```diff
@@ -167,6 +167,14 @@ public function getType()
             : null;
     }
 
+    /**
+     * @return ?TypeGenerator
+     */
+    public function getTypeObject()
+    {
+        return $this->type;
+    }
```
Ocramius: Exposing the type to the outside world seems like a regression here, as it will reduce the ability to change the codebase later on.

src/Generator/PromotedParameterGenerator.php
```diff
@@ -87,6 +87,14 @@ public static function fromParameterGeneratorWithVisibility(ParameterGenerator $
             );
         }
 
+        if (
+            null !== $type
+            && ($typeObject = $generator->getTypeObject())
+            && $typeObject->getNullable()
+        ) {
+            $type = '?' . $type;
+        }
```
Ocramius: Wondering if the `$type` of `ParameterGenerator` could be exposed internally only: we could then get rid of the string type overall

Grundik: Changed that to exposing only nullable flag.

Ocramius: @Grundik I don't have a clear mind on the approach here yet, but I'm trying to think of a design that removes `public`, and ideally avoids `protected` (although the inheritance design is already out of the bag)

Grundik: The other way I see, is to make ParameterGenerator#generateTypeHint public, this would not require exposing getNullable().

But it looks like no big deal to expose that: there a lot of other flags exposed in a same way: `passedByReference`, `vadiadic` and such.

Ocramius: Right, I'm much in "damage control" mode here, since the generator, builder and reflection-alike structure has been merged together in the past, making changes to internal structures lead to major versions :) 

I'll sleep over it, but I'd rather not expose more than is strictly needed for `->generate()` to work.

Grundik: Currenly there are no simple way to get this flag. E.g. parsing result of `ParameterGenerator->generate()` would be ugly, fragile, slow and overcomplicated.

Ocramius: Nono, I'm absolutely not suggesting that we go down the parsing way :)

Grundik:
Just to be clear: this is not the fix for https://github.com/Ocramius/ProxyManager/issues/784, there are at leas two different issues:
* this one, with missing nullable flag (and seems like https://github.com/laminas/laminas-code/issues/182 also could kead to problems);
* proxy-manager one: these reflections are used in inheritance context there, hence they dont need to redeclare promoted properties at all in that particular situation.

src/Generator/PromotedParameterGenerator.php
Grundik: Any additional thoughs?..

PR #186, opened 2023-06-07
Fix generating null return type from reflection in TypeGenerator

### LINKED ISSUE #185: TypeGenerator fails with type "null" cannot be nullable
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 14.12.x

#### Summary

Consider the following method:

```php
    public function alwaysNull(): null
    {
        return null;
    }
```

When I generate a method using `MethodGenerator::copyMethodSignature()`, the `TypeGenerator` fails with:

> Type "null" cannot be nullable

robertlemke:
The TypeGenerator failed with "type 'null' cannot be nullable" when `fromReflectionType()` was called for a method with a return type "null" due to a incorrect comparison with the string "null". This change corrects the comparison so that such methods are reflected without errors.

Obvious fix.

Resolves: #185

### DIFF
--- a/src/Generator/TypeGenerator.php
+++ b/src/Generator/TypeGenerator.php
@@ -69,7 +69,7 @@ final class TypeGenerator implements GeneratorInterface
 
         return new self(
             $atomicType,
-            $atomicType->type !== 'mixed' && $atomicType !== 'null' && $type->allowsNull()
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

gsteel:
Thank you @robertlemke :)

PR #153, opened 2022-11-18
Deprecating all `::fromArray()` methods in in the package

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Deprecating all `::fromArray()` methods in in the package

These methods are brittle, badly documented, and very much
type-unsafe.

The `src/Generator` API is mostly composed of builder objects,
and `::fromArray()` is not necessary in it.

Please also do not use arbitrary user input, deserialized as
array, to generate code: it's a recipe for disaster.

PR #188, opened 2023-08-24
Use expected DEPLOY_TOKEN env for docs build workflow

Xerkus:
As noted in #153 docs are currently not deployed due to incorrect, old?, setup used in docs workflow.


froschdesign (approved):
Good catch! 👍🏻 

Reference: https://github.com/laminas/documentation-theme/tree/master/github-actions/docs#environment

PR #189, opened 2023-09-06
Allow enum value to be backed by an empty string

martin-helmich:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Currently, the `EnumGenerator` class enforces (in its Psalm typings) that a case of a backed enumeration is backed by `int|non-empty-string`. However, the `non-empty-string` constraint is unnecessarily strict, since (AFAIK) 
an empty string is an entirely valid value (not name!) for a backed enum case.

This change relaxes the Psalm type hints a bit in order to allow this:

```php
$enum = EnumGenerator::withConfig([
  "name"        => "Foo",
  "backedCases" => [
    "type"  => "string",
    "cases" => [
      "empty" => "",
      "foo"   => "foo",
      "bar"   => "bar",
    ],
  ]
]);
```

### DIFF
--- a/src/Generator/EnumGenerator/Cases/CaseFactory.php
+++ b/src/Generator/EnumGenerator/Cases/CaseFactory.php
@@ -26,7 +26,7 @@ final class CaseFactory
      *          cases: array<non-empty-string, int>,
      *      }|array{
      *          type: 'string',
-     *          cases: array<non-empty-string, non-empty-string>,
+     *          cases: array<non-empty-string, string>,
      *      },
      * } $options
      * @return BackedCases|PureCases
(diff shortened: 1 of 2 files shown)


Ocramius:
@martin-helmich CI shows that more adjustments are needed

martin-helmich:
Yep, already noticed. Converting back to draft, for now. Although I suspect that (some) of the failing CI jobs may be unrelated to this change ([this one](https://github.com/laminas/laminas-code/actions/runs/6098689159/job/16549094422?pr=189), for example).

martin-helmich:
Ready for review, again. The Psalm CI job now runs successfully; some PHPUnit jobs are failing for reasons I believe to be unrelated to this change.

Ocramius (approved):
Thanks @martin-helmich!

Issue #190, opened 2023-09-11
PHP 8.3: New features


alexander-schranz:
### Feature Request

I thought I put all things which are PHP 8.3 related into this issue.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

 - [x] Check if existing Generators work with PHP 8.3 (https://github.com/laminas/laminas-code/pull/191)
 - [ ] Typed class constants (https://wiki.php.net/rfc/typed_class_constants) (think const visibilty is also not already implemented, looks somebody already tried it but very complicated: https://github.com/laminas/laminas-code/pull/32)
 - [ ] Dynamic class constant fetch (https://wiki.php.net/rfc/dynamic_class_constant_fetch)
 - [ ] Anonymous readonly classes (https://github.com/php/php-src/blob/011071a3b3f0f60c3d8ff70479251ed8873f90ec/UPGRADING#L115) (think anonymous classes in general do not exist yet)
 
Based on https://stitcher.io/blog/new-in-php-83 I think this are the only features which could have effects and could make sense to add to laminas-code generators.

PR #191, opened 2023-09-11
Add PHP 8.3 to supported PHP version

alexander-schranz:
|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Requires:

 - [ ] PHP 8.3 Support laminas-stdlib: https://github.com/laminas/laminas-stdlib/commit/8ec8790aed1d0aead6d0155ff8ba15a6d06810c4 (merged not yet released)

### Description

This will just make sure if the current supported generators still work as expected in PHP 8.3. The features provided by PHP 8.3 can be build then ontop of this. A list of features which could be tackled by somebody can be found in the following issue: https://github.com/laminas/laminas-code/issues/190

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "~8.1.0 || ~8.2.0"
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0"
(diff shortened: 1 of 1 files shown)


Ocramius:
Handled in #194

PR #192, opened 2023-09-11
CI fixes

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Unit tests are failing because PHPUnit needs pinning to a more recent release and the config file needs updating as per schema changes.

Reset the Psalm baseline - Expanded with new issues.

Other minor fixes to the README and various dot files.

This should get CI green so contributors aren't getting messed around with issues that have nothing to do with their patch.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,0 @@
-coverage_clover: clover.xml
-json_path: coveralls-upload.json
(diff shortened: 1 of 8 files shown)


Ocramius (approved):
Thanks @gsteel!

PR #200, opened 2024-06-17
Return static instance on abstract class setters

veewee:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no/yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR improves static analysis when using the setters that are located in the abstract generators:
Instead of returning with type `Abstract*` on setters, the `static` instance will be returned.
This makes static analyzers understand a fluent approach better.

For example:

```php
(new MethodGenerator('getX'))
    ->setStatic(true)
    // > This would previously result in the static analyzer to believe the type is `AbstractMemberGenerator`
    ->setBody('return '.$code)
    ->setReturnType(ClassMapCollection::class)
```

Resulting in issues like:

```
>   54     Parameter #1 $method of method
         Laminas\Code\Generator\ClassGenerator::addMethodFromGenerator()
         expects Laminas\Code\Generator\MethodGenerator,
         Laminas\Code\Generator\AbstractMemberGenerator given.
```

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -37,10 +37,10 @@
   </file>
   <file src="src/Generator/AbstractMemberGenerator.php">
     <PossiblyUnusedReturnValue>
-      <code>AbstractMemberGenerator</code>
-      <code>AbstractMemberGenerator</code>
-      <code>AbstractMemberGenerator</code>
-      <code>AbstractMemberGenerator</code>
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
Thanks @veewee

PR #201, opened 2024-10-01
Add Support for PHP 8.4

nishant04412:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
         "ext-phar": "*",
         "doctrine/annotations": "^2.0.1",
         "laminas/laminas-coding-standard": "^2.5.0",
-        "laminas/laminas-stdlib": "^3.17.0",
-        "phpunit/phpunit": "^10.3.3",
+        "laminas/laminas-stdlib": "^3.18.0",
+        "phpunit/phpunit": "^10.4.2",
         "psalm/plugin-phpunit": "^0.19.0",
         "vimeo/psalm": "^5.15.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "doctrine/annotations": "^2.0.1",
         "laminas/laminas-coding-standard": "^2.5.0",
         "laminas/laminas-stdlib": "^3.18.0",
-        "phpunit/phpunit": "^10.4.2",
+        "phpunit/phpunit": "^10.5.37",
         "psalm/plugin-phpunit": "^0.19.0",
         "vimeo/psalm": "^5.15.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0"
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0"
     },
     "require-dev": {
         "ext-phar": "*",
--- a/test/Generator/TestAsset/ParameterClass.php
+++ b/test/Generator/TestAsset/ParameterClass.php
@@ -75,7 +75,7 @@ class ParameterClass
 
     }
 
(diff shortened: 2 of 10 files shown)


W0rma:
Seems like the PHP 8.4 pipeline is blocked by https://github.com/laminas/laminas-stdlib/pull/118

nishant04412:
Yes and https://github.com/laminas/laminas-stdlib/pull/118 looks to be blocked due to https://github.com/vimeo/psalm/pull/10928

I added a comment https://github.com/laminas/laminas-stdlib/pull/118#issuecomment-2384861365 to confirm same. Lets hope dependent PRs get merged soon to unblock this.

nishant04412:
@gsteel I am facing some issues while resolving one of the issue in https://github.com/laminas/laminas-code/blob/562e02b7d85cb9142b5116cc76c4c7c162a11a1c/test/Generator/TestAsset/ParameterClass.php#L78

Here, if we add `?` before `\stdClass $a = null` then it'll resolve PHP 8.4 related issue but then test is giving deprecation `Optional parameter $a declared before required parameter $b is implicitly treated as a required parameter`. 

Please suggest if we can keep mandatory parameter $b before optional parameter $a in this ? Or any other advise if not.

gsteel:
> Please suggest if we can keep mandatory parameter $b before optional parameter $a in this ? Or any other advise if not.

`$a` is effectively mandatory because `$b` is and we don't support BC on named parameters, so the signature can be changed to `fn(stdClass|null $a, $b)` without breaking BC AFAIC. Alternatively, retain the `= null` and just suppress the CS issue.

Also, this is a Test asset - BC is not a concern.

samsonasik:
default `null` value can be removed if next parameter is required, see

https://3v4l.org/4pQLN

that safe even extended, also, named argument is not used in the test :)

test/Generator/TestAsset/ParameterClass.php
```diff
@@ -75,7 +75,7 @@ public function defaultConstant($con = self::FOO)
 
     }
 
-    public function defaultObjectEqualsNullAndNotOptional(\stdClass $a = null, $b)
+    public function defaultObjectEqualsNullAndNotOptional(?\stdClass $a, $b)
```
samsonasik: hm..., this method name seems no longer represent to display, probably just rename the method

nishant04412: Okay. How does name below name looks like ?

`public function handleNullableObject(?\stdClass $a, $b)`

samsonasik: `nullableAndRequired` is fine for me

composer.json
```diff
@@ -9,14 +9,14 @@
...
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0"
     },
     "require-dev": {
         "ext-phar": "*",
         "doctrine/annotations": "^2.0.1",
         "laminas/laminas-coding-standard": "^2.5.0",
-        "laminas/laminas-stdlib": "^3.17.0",
-        "phpunit/phpunit": "^10.3.3",
+        "laminas/laminas-stdlib": "^3.18.0",
+        "phpunit/phpunit": "^10.4.2",
```
samsonasik: update to latest phpunit `^10.5.37` to fix lowest issue on php 8.4

```php
        "phpunit/phpunit": "^10.5.37",
```

then, regenerate composer.lock

nishant04412:
@samsonasik Please review and let me know if clarification or code modification required.

samsonasik:
Sign off is needed, read https://github.com/laminas/laminas-code/pull/201/checks?check_run_id=32024385175

samsonasik (approved):

nishant04412:
@gsteel PR changes are approved. Can you please check and merge this PR ?

gsteel (approved):
Thank you @nishant04412

PR #207, opened 2025-02-27
Fix PropertyTag::setPropertyName() deprecation warning

lifinsky:
```
ErrorException: Deprecated: ltrim(): Passing null to parameter #1 ($string) of type string is deprecated
#18 /vendor/laminas/laminas-code/src/Generator/DocBlock/Tag/ParamTag.php(55): Laminas\Code\Generator\DocBlock\Tag\ParamTag::setVariableName
#17 /vendor/laminas/laminas-code/src/Generator/DocBlock/TagManager.php(59): Laminas\Code\Generator\DocBlock\TagManager::createTagFromReflection
#16 /vendor/laminas/laminas-code/src/Generator/DocBlockGenerator.php(48): Laminas\Code\Generator\DocBlockGenerator::fromReflection
#15 /vendor/laminas/laminas-code/src/Generator/MethodGenerator.php(47): Laminas\Code\Generator\MethodGenerator::fromReflection
```

### DIFF
--- a/src/Generator/DocBlock/Tag/PropertyTag.php
+++ b/src/Generator/DocBlock/Tag/PropertyTag.php
@@ -37,7 +37,7 @@ class PropertyTag extends AbstractTypeableTag implements TagInterface
      */
     public function setPropertyName($propertyName)
     {
-        $this->propertyName = ltrim((string) $propertyName, '$');
+        $this->propertyName = ! empty($propertyName) ? ltrim($propertyName, '$') : '';
         return $this;
     }
 
--- a/src/Generator/DocBlock/Tag/PropertyTag.php
+++ b/src/Generator/DocBlock/Tag/PropertyTag.php
@@ -37,7 +37,7 @@ class PropertyTag extends AbstractTypeableTag implements TagInterface
      */
     public function setPropertyName($propertyName)
     {
-        $this->propertyName = $propertyName ? ltrim($propertyName, '$') : null;
+        $this->propertyName = ltrim((string) $propertyName, '$');
         return $this;
     }
 
--- a/src/Generator/DocBlock/Tag/PropertyTag.php
+++ b/src/Generator/DocBlock/Tag/PropertyTag.php
@@ -37,7 +37,7 @@ class PropertyTag extends AbstractTypeableTag implements TagInterface
      */
     public function setPropertyName($propertyName)
     {
-        $this->propertyName = ltrim($propertyName, '$');
+        $this->propertyName = $propertyName ? ltrim($propertyName, '$') : null;
         return $this;
     }
 


lifinsky:
@alexmerlin

dgafka:
@Ocramius are you able to help to push that forward?

lifinsky:
@gsteel

src/Generator/DocBlock/Tag/PropertyTag.php
```diff
@@ -37,7 +37,7 @@ public function getName()
      */
     public function setPropertyName($propertyName)
     {
-        $this->propertyName = ltrim($propertyName, '$');
+        $this->propertyName = $propertyName ? ltrim($propertyName, '$') : null;
```
samsonasik: use directly cast string `ltrim((string) $propertyName, '$')` instead, ltrim never return null. https://3v4l.org/ufoMl

lifinsky:
```php
/**
     * @return string|null
     */
    public function getPropertyName()
    {
        return $this->propertyName;
    }
```

Empty string should be transformed to null. Why do we need cast null to string?)

gsteel: This should be annotated as `/** @param non-empty-string $propertyName */` and the string asserted as non-empty in the method body with `assert(is_string($propertyName) && $propertyName !== '')`, or prefereably `Assert::stringNotEmpty($propertyName)` if an assertion library is used here.

If the method is being called with `null`, the caller is wrong and it should crash, or at least have an assertion fail in tests.

Adding `non-empty-string` as the documented type _and_ asserting it in the method body will cause `DocBlockTypeContradiction` in psalm that can be baselined.

How is this even receiving `null` when it is documented as `string`?

Is this deprecation triggered in the tests here? or in consuming code?

lifinsky: In the code using this library, it’s possible that the `setPropertyName` method shouldn’t have been exposed in a higher-level class at all, but the goal of this PR was to quickly resolve the issue with minimal risk.

gsteel (changes requested):
@lifinsky `null` is not an acceptable parameter here.

Simply casting with `(string)` because it _might_ receive a null argument introduces the probability that someone will send an array and we'll have a PHP warning for array to string conversion.

This method should either defend against invalid parameters with assertions, or do nothing _(as the caller is simply using it wrong)_ and wait for someone to refactor this class to be an immutable object with strict type guarantees.

lifinsky:
> @lifinsky `null` is not an acceptable parameter here.
> 
> Simply casting with `(string)` because it _might_ receive a null argument introduces the probability that someone will send an array and we'll have a PHP warning for array to string conversion.

Done

lifinsky:
@gsteel The linter didn’t allow casting to (string) because it’s redundant — PHPDoc already defines it as a string. I replaced it with !empty, as done elsewhere.

src/Generator/DocBlock/Tag/PropertyTag.php
```diff
@@ -37,7 +37,7 @@ public function getName()
      */
     public function setPropertyName($propertyName)
     {
-        $this->propertyName = ltrim($propertyName, '$');
+        $this->propertyName = ! empty($propertyName) ? ltrim($propertyName, '$') : '';
```
samsonasik, line 40:
```suggestion
        $this->propertyName = ltrim((string) $propertyName, '$');
```

samsonasik:
@gsteel use cast is needed to keep existing behaviour, except it needs to go to 4.0.0

src/Generator/DocBlock/Tag/PropertyTag.php
lifinsky: @samsonasik linter does not allow this. Please approve my version.

lifinsky: As you can see other methods also have this expression.

lifinsky: https://github.com/laminas/laminas-code/pull/207#issuecomment-3429629936

samsonasik: I think the issue is why user pass null on the exact signature string, which should never happen, while docblock, should be followed. The notice is expected as it not follow the signature, the refactor then imo should be on userland.

lifinsky: > I think the issue is why user pass null on the exact signature string, which should never happen, while docblock, should be followed. The notice is expected as it not follow the signature, the refactor then imo should be on userland.

The stack trace was provided above.
This may occur due to incomplete PHPDoc.
In any case, it’s an issue within one of the library’s classes.
The current solution only suppresses the notice and doesn’t break anything internally.
It’s clear that the library needs stricter type control by using strict_types.

samsonasik: That's nothing todo with strict_types. That's php feature of deprecating `null` support on string functions on php 8.1+.

nullable property is related with initialization, while filling its value should follow the doc as a signature.

The empty value already checked on construct via empty, when filling via setter, exact string per signature should be followed.

samsonasik:
I am going to close this, this is expected, as signature is string, which not allow null, even docblock should be respected by the app that uses it.

The refactor will require major version.

Thank you for understanding.

src/Generator/DocBlock/Tag/PropertyTag.php
lifinsky: @dgafka maybe we should fix our "client" side is it possible

PR #208, opened 2025-07-19
Remove extra space before return‑type colon in `MethodGenerator`

alexchexes:
PSR‑12 and Laminas CS v2/v3 specify **no space** before the `:` in return‑type declarations (`function foo(): bool`).
But `MethodGenerator::generate()` still outputs `) : bool`, a legacy Zend CS formatting.  
This patch drops the extra space, bringing generated code in line with current standards.
Behaviour is unchanged; all affected tests are updated and passing.

|    Q          |   A
|-------------- | ------
| QA            | yes

* **What changed?**  
  Replaced `' : '` with `': '` in `MethodGenerator::generate()`.

* **Why?**  
  To comply with PSR‑12 / Laminas CS v2 return‑type spacing.

* **Impact**  
  Cosmetic only; no functional changes. Test suite green.

### DIFF
--- a/src/Generator/MethodGenerator.php
+++ b/src/Generator/MethodGenerator.php
@@ -376,7 +376,7 @@ class MethodGenerator extends AbstractMemberGenerator implements Stringable
         $output .= ')';
 
         if ($this->returnType) {
-            $output .= ' : ' . $this->returnType->generate();
+            $output .= ': ' . $this->returnType->generate();
         }
 
         if ($this->isAbstract()) {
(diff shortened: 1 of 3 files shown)


gsteel (approved):
LGTM! Thanks @alexchexes 

That said, I'll defer to @Ocramius for the final word - cosmetics like this should probably be handled by PHP_CodeSniffer or CS Fixer

alexchexes:
Thanks! 

> cosmetics like this should probably be handled by PHP_CodeSniffer or CS Fixer

Having the generator emit PSR‑12‑compliant code up front means even projects that don't run CS fixers get the right style for free, so it would be great if we eventually merge this.

Unlike some other codestyle preferences, which may even vary from project to project, an extra space in the `fn() : bool` is a rare style choice today and catches the eye too easily, IMO

PR #211, opened 2025-10-31
Add support for PHP 8.5, Remove support for PHP 8.1

acelaya:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR adds support for PHP 8.5, and drops support for PHP 8.1

### DIFF
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -1,5 +1,6 @@
 {
   "ignore_php_platform_requirements": {
-    "8.4": true
+    "8.4": true,
+    "8.5": true
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "doctrine/annotations": "^2.0.1",
         "laminas/laminas-coding-standard": "^3.0.0",
         "laminas/laminas-stdlib": "^3.18.0",
-        "phpunit/phpunit": "^10.5.37",
+        "phpunit/phpunit": "^10.5.58",
         "psalm/plugin-phpunit": "^0.19.0",
         "vimeo/psalm": "^5.15.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0"
+        "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0"
     },
     "require-dev": {
         "ext-phar": "*",
@@ -29,7 +29,7 @@
             "dealerdirect/phpcodesniffer-composer-installer": true
         },
         "platform": {
-            "php": "8.1.99"
+            "php": "8.2.99"
         },
         "sort-packages": true
     },
--- a/test/Reflection/ParameterReflectionTest.php
+++ b/test/Reflection/ParameterReflectionTest.php
@@ -14,6 +14,8 @@ use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\TestCase;
 use ReflectionType;
 
+use const PHP_VERSION_ID;
+
 #[Group('Laminas_Reflection')]
 #[Group('Laminas_Reflection_Parameter')]
 class ParameterReflectionTest extends TestCase
@@ -159,7 +161,13 @@ class ParameterReflectionTest extends TestCase
             [InternalHintsClass::class, 'floatParameter', 'foo', 'float'],
             [InternalHintsClass::class, 'stringParameter', 'foo', 'string'],
             [InternalHintsClass::class, 'boolParameter', 'foo', 'bool'],
-            [ClassTypeHintedClass::class, 'selfParameter', 'foo', 'self'],
+            [
+                ClassTypeHintedClass::class,
+                'selfParameter',
+                'foo',
+                // In PHP >=8.5, `ReflectionType::getName()` does not return 'self'
+                PHP_VERSION_ID >= 80500 ? ClassTypeHintedClass::class : 'self',
+            ],
             [ClassTypeHintedClass::class, 'classParameter', 'foo', ClassTypeHintedClass::class],
             [ClassTypeHintedClass::class, 'otherClassParameter', 'foo', InternalHintsClass::class],
             [ClassTypeHintedClass::class, 'closureParameter', 'foo', Closure::class],
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "92b57f70bfe1072d628636e36fbf4e7a",
+    "content-hash": "abc68b3c2226ea90a6785823e6019b55",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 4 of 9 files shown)


.laminas-ci.json
```diff
@@ -1,5 +1,6 @@
 {
   "ignore_php_platform_requirements": {
-    "8.4": true
+    "8.4": true,
+    "8.5": true
```
acelaya, line 4: I meant to replace `8.4` with `8.5` here, but latest psalm 5 does not seem to be compatible with PHP 8.4 either.

I don't think updating to psalm 6 is something that should be done in this PR, but rather separately, so I kept both 8.4 and 8.5 here.

acelaya:
I'm looking into this failing test:

```diff
1) LaminasTest\Code\Reflection\ParameterReflectionTest::testGetType with data set #6 ('LaminasTest\Code\TestAsset\Cl...dClass', 'selfParameter', 'foo', 'self')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'self'
+'LaminasTest\Code\TestAsset\ClassTypeHintedClass'
```

composer.json
```diff
@@ -9,7 +9,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0"
+        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0"
```
gsteel: Please drop support for 8.1 whilst you're here 👍

gsteel (changes requested):
Looks like `ReflectionType::getName()` has changed - it's not documented though: https://github.com/php/php-src/blob/php-8.5.0RC3/UPGRADING

composer.json
gsteel: Forgot to mention that `config.platform.php` will need to be set to `8.2.99`

acelaya: Oops, I didn't realize that was set.

acelaya:
> Looks like `ReflectionType::getName()` has changed - it's not documented though: https://github.com/php/php-src/blob/php-8.5.0RC3/UPGRADING

How would you suggest proceeding here? Considering the same test would now require different assertions for PHP 8.5 and previous versions?

alexislefebvre:
> > Looks like `ReflectionType::getName()` has changed - it's not documented though: https://github.com/php/php-src/blob/php-8.5.0RC3/UPGRADING
> 
> How would you suggest proceeding here? Considering the same test would now require different assertions for PHP 8.5 and previous versions?

There are several ways:

- use `version_compare()` to split the different assertions: https://www.php.net/manual/en/function.version-compare.php#example-685
- write 2 different tests and use PHPUnit’s `RequiresPhp`: https://docs.phpunit.de/en/12.4/attributes.html#requiresphp to target `< 8.5` and `>= 8.5`. Older versions of PHPUnit provide it through an annotation: https://docs.phpunit.de/en/11.4/annotations.html#requires

test/Reflection/ParameterReflectionTest.php
```diff
@@ -159,7 +163,13 @@ public static function reflectionHints()
             [InternalHintsClass::class, 'floatParameter', 'foo', 'float'],
             [InternalHintsClass::class, 'stringParameter', 'foo', 'string'],
             [InternalHintsClass::class, 'boolParameter', 'foo', 'bool'],
-            [ClassTypeHintedClass::class, 'selfParameter', 'foo', 'self'],
+            [
+                ClassTypeHintedClass::class,
+                'selfParameter',
+                'foo',
+                // In PHP >=8.5, `ReflectionType::getName()` does not return 'self'
+                version_compare(PHP_VERSION, '8.5.0', '<') ? 'self' : ClassTypeHintedClass::class,
```
gsteel: Don't you need `ge` or `>=` here?

gsteel:
@acelaya - If you can also upgrade PHPUnit to `11.x` here too, we'll lose a load of deprecations too. Normally, I'd suggest also adding `failOnDeprecation` and `failOnPhpUnitDeprecation` to `phpunit.xml`, but only if the changes required are minimal - we can drop all calls to `ReflectionProperty::setAccessible()` because it's a no-op since 8.2

test/Reflection/ParameterReflectionTest.php
samsonasik: use proper `PHP_VERSION_ID` constant comparison with `80500` is better imo that self explanatory

acelaya: > Don't you need ge or >= here?

Actually, I think it's right this way, unless I'm not understanding the error PHPUnit is reporting 🤔 

On the other hand, this did not change anything, which is weird. If the condition was in fact flipped, it should have made the test fail for the rest of the PHP versions as well.

I'm trying to get a local PHP 8.5 docker container set to run the tests instead of relying on GitHub actions, so that I can get a faster feedback cycle and avoid spamming everyone 😅 

> use proper `PHP_VERSION_ID` constant comparison with `80500` is better imo that self explanatory

I'm not sure if I follow what you are suggesting 🤔

samsonasik: Reading `version_compare()` need to understand what right and left actually means, while compare, eg: `PHP_VERSION_ID >= 80500` is easier to understand by directly look the code.

acelaya: Gotcha.

Yeah, I totally agree. I did have to check how `version_compare` works before using it here 😅

alexisLefebvre: > > use proper `PHP_VERSION_ID` constant comparison with `80500` is better imo that self explanatory
> 
> I'm not sure if I follow what you are suggesting 🤔

It's another way without `version_compare()`:

```php
if (\PHP_VERSION_ID >= 80500) {
```

Example: https://github.com/symfony/symfony/blob/8a58816cc8fd78732d878e20c975dc64e09dca86/src/Symfony/Component/VarDumper/Caster/SocketCaster.php#L29

alexislefebvre:
> @acelaya - If you can also upgrade PHPUnit to `11.x` here too, we'll lose a load of deprecations too. Normally, I'd suggest also adding `failOnDeprecation` and `failOnPhpUnitDeprecation` to `phpunit.xml`, but only if the changes required are minimal - we can drop all calls to `ReflectionProperty::setAccessible()` because it's a no-op since 8.2

What about doing this in a separate PR? It may be easier to treat errors due to the PHUnit and PHP 8.5 upgrades separately.

acelaya:
> @acelaya - If you can also upgrade PHPUnit to `11.x` here too, we'll lose a load of deprecations too. Normally, I'd suggest also adding `failOnDeprecation` and `failOnPhpUnitDeprecation` to `phpunit.xml`, but only if the changes required are minimal - we can drop all calls to `ReflectionProperty::setAccessible()` because it's a no-op since 8.2

I'm trying to do this, but it seems it might require updating psalm to v6 first.

The problem is that psalm 5 requires `sebastian/diff: ^4.0 || ^5.0`, while phpunit 11 requires `sebastian/diff: ^6.0`.

Would you be ok separating the update to PHPUnit 11 out of this PR, until the project has been updated to psalm 6? The only problem is that the deprecation warnings are making the tests be marked as failed, even though they are actually passing now.

EDIT: Actually, I could simply remove those `setAccessible(...)` calls. That should solve the deprecation warnings at least.

composer.json
```diff
@@ -9,14 +9,14 @@
...
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0"
+        "php": "~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0"
     },
     "require-dev": {
         "ext-phar": "*",
         "doctrine/annotations": "^2.0.1",
         "laminas/laminas-coding-standard": "^3.0.0",
         "laminas/laminas-stdlib": "^3.18.0",
-        "phpunit/phpunit": "^10.5.37",
+        "phpunit/phpunit": "^10.5.58",
```
acelaya, line 19: I update to the latest PHPUnit 10.5 patch to solve a deprecation warning triggered with PHP 8.5 when using lowest dependencies.

acelaya:
Ok, CI is green now. I did the minimum changes to get it working.

As I mentioned [here](https://github.com/laminas/laminas-code/pull/211#issuecomment-3475943144), I could not update to PHPUnit 11 as it has a shared dependency with psalm, but with no overlapping in version constraints. The project probably needs to update to psalm 6 first.

gsteel (approved):
Thanks @acelaya 👍

